PR #3, opened 2021-07-02
Fix the to_snake_case function

kylestach:
The previous implementation had issues with message names like MultiDOFJointState.msg, which it converted to multi_dofjoint_state.msg. This would cause a mismatch in generated and expected output names.

### DIFF
--- a/ros2/interfaces.bzl
+++ b/ros2/interfaces.bzl
@@ -53,45 +53,34 @@ ros2_interface_library = rule(
     implementation = _ros2_interface_library_impl,
 )
 
-def to_snake_case(not_snake_case):
+def to_snake_case(value):
     """ Converts camel-case to snake-case.
 
-    Based on https://stackoverflow.com/a/19940888.
+    Based on convert_camel_case_to_lower_case_underscore from rosidl_cmake.
+    Unfortunately regex doesn't exist in Bazel.
 
     Args:
       not_snake_case: a camel-case string.
     Returns:
       A snake-case string.
     """
-    if not_snake_case.isupper():
-        return not_snake_case.lower()
+    result = ""
+    value_padded = ' ' + value + ' '
+    for i in range(len(value)):
+        prev_char = value_padded[i]
+        char = value_padded[i + 1]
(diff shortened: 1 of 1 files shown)


ros2/interfaces.bzl
```diff
@@ -56,42 +56,29 @@ ros2_interface_library = rule(
...
-                final += item.lower()
-            else:
-                final += "_" + item.lower()
-        else:
-            final += item
-    if final[0] == "_":
-        final = final[1:]
-    return final
+    result = ""
+    not_snake_case_padded = ' ' + not_snake_case + ' '
```
mvukov, line 68: Here and below, please use double quotes, that's somewhat usual when writing Starlark code.

ros2/interfaces.bzl
```diff
@@ -56,42 +56,29 @@ ros2_interface_library = rule(
...
+    for i in range(len(not_snake_case)):
+        prev_char, char, next_char = not_snake_case_padded[i:i+3].elems()
+        # insert an underscore before any upper case letter
+        # which is not followed by another upper case letter
+        if char.isupper() and next_char.islower() and prev_char != ' ':
+            result += '_'
+        # insert an underscore before any upper case letter
+        # which is preseded by a lower case letter or number
+        elif char.isupper() and (prev_char.islower() or prev_char.isdigit()):
+            result += '_'
```
mvukov, line 78: IMO a bit easier to follow is to embed the comments within if/elif:
```Starlark
if char.isupper() and next_char.islower() and prev_char != ' ':
    # Insert an underscore before any upper case letter
    # which is not followed by another upper case letter.
    result += '_'
elif char.isupper() and (prev_char.islower() or prev_char.isdigit()):
    # Insert an underscore before any upper case letter
    # which is preceded by a lower case letter or number.
    result += '_'
```
I also fixed formatting of sentences here.

mvukov (commented):
Great, many thanks! I still don't understand why simple 1-1 file name mapping wasn't good enough for ROS2. Perhaps that's something to be rectified in ROS3 :)

I left some minor comments related to formatting and readability.

PR #6, opened 2022-05-12
Added support for sensor_msgs

Sanjaer:
I added support for sensor_msgs as I needed it for my project


mvukov (approved):
Many thanks! :100:

PR #12, opened 2022-10-06
Use versioned repo names to avoid conflict

LiyouZhou:
When importing rules_ros2 into another project `fmt` and `libyaml` causes name conflict. This patch uses versioned names to avoid chances of conflict.

### DIFF
--- a/repositories/repositories.bzl
+++ b/repositories/repositories.bzl
@@ -93,7 +93,7 @@ def ros2_repositories():
 
     maybe(
         http_archive,
-        name = "fmt",
+        name = "fmt-7.1.3",
         build_file = "@com_github_mvukov_rules_ros2//repositories:fmt.BUILD.bazel",
         sha256 = "5cae7072042b3043e12d53d50ef404bbb76949dad1de368d7f993a15c8c05ecc",
         strip_prefix = "fmt-7.1.3",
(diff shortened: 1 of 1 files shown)


mvukov:
I intentionally defined `http_archive` targets under `maybe` macro such that overrides is possible. That said, if another fmt, libyaml, etc. are declared before the call to
```Python
load("//repositories:repositories.bzl", "ros2_repositories")

ros2_repositories()  # This one in particular
```
your project, rules_ros2 as well, are going to use those ftm and libyaml versions. I am curious to learn how your workspace setup looks like. Can you provide more info on the problems that you're facing?

mvukov:
(I updated CI config and reopened the PR to trigger CI checks on your PR.)

LiyouZhou:
We have monorepo and there are global 3rd party imports for `fmt` and `libyaml` already (for other part of the codebase unrelated to the robot), unfortunately the existing 3rd party imports appear to be incompatible versions. Hence to avoid further conflict, I want to specify the version exactly and use those solely for rules_ros2.

LiyouZhou:
@mvukov do you think this can be merged?

mvukov:
OK, some parts of your monorepo import (before rules_ros2 is imported) fmt&libyaml that don't work with rules_ros2. For that reason you want separate versions of those packages just for rules_ros2. That's OK, but, there is a practical chance that a dev can use a library that depends on fmt version x with a ros node that depends on fmt version y with your PR ---> linker would freak out, I think. I guess you prevent those clashes with proper target visibility settings. The [shadowing deps](https://docs.bazel.build/versions/5.1.0/external.html#shadowing-dependencies) approach you use is the way to go for you. I think it's a good solution for your use-case, but, not a generic one. For this reason, I am inclined not to merge this in the main branch. One day when this repo has a troubleshooting section, this PR should be referenced.

PR #13, opened 2022-10-06
Fix python import issue

LiyouZhou:
Traditionally bazel created init files in the root of the build folder of python packages, this causes confusion during python import when the build folder name is the same as the package name. See this [discussion](https://github.com/bazelbuild/bazel/issues/12238). So you end up with:

```
- my_package
  - __init__.py <==== created by bazel
  - my_package <==== copy of the source code
     - __init__.py <==== from the actual source code
```

`legacy_create_init=False` will prevent this `__init__.py` from being created.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "cmake.configureOnOpen": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "cmake.configureOnOpen": false
-}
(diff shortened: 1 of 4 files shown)


mvukov:
Thanks for the contribution. Which Bazel version are you using btw?

I use 5.2 on daily basis and until yesterday I haven't had any problems using `--incompatible_default_to_explicit_init_py` option (which is also in .bazelrc of this repo and I hope you use it as well). And then I found a Bazel bug, see https://github.com/bazelbuild/bazel/issues/16388. Looks like you have the same issue here. Extra flags won't harm. LGTM!

LiyouZhou:
Thx, currently `bazel 5.3.1`.

PR #16, opened 2022-10-14
Upgrade to humble

LiyouZhou:
- Upgrade all packages to humble
- add lifecycle api
- tested cpp examples to work

@mvukov I didn't work on the rclpy package. Would you be able to help?


mvukov (commented):
This is super cool, many thanks for digging into this.

Please take a look at the latest main. I now use official ros2.repos files from github.com/ros2/ros2 to auto-generate http_archive targets for ROS2 packages, see repositories/private/BUILD.bazel. And I also started using patches like you do in this PR :+1: .

I'll definitely help with rclpy, and possibly other stuff along the way. I just need a bit of time to wrap my head around this.

Regarding cyclonedds: I am really considering to use rules_foreign_cc + it's cmake extension here instead of maintaining this complicated BUILD file. WDYT? From what I know, it should work just fine with cross-compilation.

mvukov:
If you have some time, please merge the latest main and resolve merge conflicts.

LiyouZhou:
Merged with main, tested with examples/chatter and examples/actions, all cpp targets work as expected.

LiyouZhou:
Re cyclone: in my experience cmake targets work surprising well, but because this extra layer of dark magic, bit difficult to debug any build issues. I will contribute another patch after this one is merged to introduce iceoryx backend for cyclone and I used cmake target for that.

Interestingly [iceoryx](https://github.com/eclipse-iceoryx/iceoryx) is switching to a bazel build system I wonder if cyclonedds will follow suite. We should raise an issue in their github repo :blush:.

LiyouZhou:
Hi @mvukov do you think this PR can be merged now?

examples/chatter/talker.cc
```diff
@@ -25,12 +25,8 @@
...
  public:
   MinimalPublisher() : Node("minimal_publisher"), count_(0) {
-    int callback_period_ms = 500;
-    declare_parameter("callback_period_ms");
-    try {
-      callback_period_ms = get_parameter("callback_period_ms").as_int();
-    } catch (const rclcpp::exceptions::ParameterNotDeclaredException&) {
-    }
+    declare_parameter("callback_period_ms", 500);
+    auto callback_period_ms = get_parameter("callback_period_ms").as_int();
```
mvukov, line 29: In foxy this fires an exception if the node is not started with the param defined. Is this different in Huble?

The idea was to have this by default set to 0.5s and set to something much smaller for tests to take less time.

repositories/patches/rclcpp-16.0.1.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/rclcpp/src/rclcpp/typesupport_helpers.cpp b/rclcpp/src/rclcpp/typesupport_helpers.cpp
```
mvukov, line 1: Regarding the patch names: there is no need to have version names therein. If we ever want to support multiple ROS2 distros, that is probably going to be done via different git branches.

repositories/patches/rclcpp-16.0.1.patch
```diff
@@ -0,0 +1,25 @@
...
+-#include "ament_index_cpp/get_package_prefix.hpp"
+-#include "ament_index_cpp/get_resources.hpp"
+ #include "rcpputils/shared_library.hpp"
+ #include "rcpputils/find_library.hpp"
+ #include "rosidl_runtime_cpp/message_type_support_decl.hpp"
+@@ -49,11 +47,6 @@ std::string get_typesupport_library_path(
+ #endif
+
+   std::string package_prefix;
+-  try {
```
mvukov, line 18: Very nice :+1:

repositories/patches/rmw_implementation_library_path.patch
```diff
@@ -1,41 +1,58 @@
...
++#else
 +  const std::string library_path = RMW_LIBRARY_PATH;
++  try {
++    return std::make_shared<rcpputils::SharedLibrary>(library_path.c_str());
++  } catch (const std::exception & e) {
++    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
++      "failed to load shared library '%s' due to %s",
++      library_path.c_str(), e.what());
++    return nullptr;
++  }
```
mvukov, line 53: In fact, we just need this part. I don't think we need `#ifndef RMW_LIBRARY_PATH` stuff.

repositories/patches/rosidl_rm_unnecessary_asserts.patch
```diff
@@ -1,72 +1,36 @@
...
      assert input_file.suffix == '.action'
  
-     abs_input_file = package_dir / input_file
--    print(f'Reading input file: {abs_input_file}')
-     abs_input_file = package_dir / input_file
-     content = abs_input_file.read_text(encoding='utf-8')
-     action = parse_action_string(package_name, input_file.stem, content)
- 
-     output_file = output_dir / input_file.with_suffix('.idl').name
-     abs_output_file = output_file.absolute()
```
mvukov, line 20: It's a bit strange that you had to remove this much code. :)

repositories/rcl.BUILD.bazel
```diff
@@ -55,3 +56,24 @@ ros2_c_library(
         "@ros2_rosidl//:rosidl_runtime_c",
     ],
 )
+
+ros2_c_library(
```
mvukov, line 60: AFAICS, this is a new library not strictly necessary for migration to humble. WDYT?

repositories/rcl_logging.BUILD.bazel
```diff
@@ -1,15 +1,27 @@
 """ Builds rcl_logging.
 """
 
-load("@com_github_mvukov_rules_ros2//ros2:cc_defs.bzl", "ros2_cpp_library")
+load("@com_github_mvukov_rules_ros2//ros2:cc_defs.bzl", "CPP_COPTS", "ros2_cpp_library")
+
+ros2_cpp_library(
```
mvukov, line 6: ros2_c_library

repositories/repositories.bzl
```diff
@@ -49,18 +49,27 @@ def ros2_repositories():
         http_archive,
         name = "libyaml",
         build_file = "@com_github_mvukov_rules_ros2//repositories:libyaml.BUILD.bazel",
-        sha256 = "2c103fc473e904c6fe7580277f1fa16b6d716e54d5e3f32a8913c4850ae03b3f",
-        strip_prefix = "libyaml-acd6f6f014c25e46363e718381e0b35205df2d83",
-        urls = ["https://github.com/yaml/libyaml/archive/acd6f6f014c25e46363e718381e0b35205df2d83.tar.gz"],
+        sha256 = "fa240dbf262be053f3898006d502d514936c818e422afdcf33921c63bed9bf2e",
+        strip_prefix = "libyaml-0.2.5",
```
mvukov, line 53: The commit that is already used is newer than this release. Can we stick to that one?

repositories/repositories.bzl
```diff
@@ -49,18 +49,27 @@ def ros2_repositories():
...
+        urls = ["https://github.com/yaml/libyaml/archive/0.2.5.tar.gz"],
+    )
+
+    maybe(
```
mvukov, line 57: Please rm this definition, it should be used from the auto-gen ros2_repositories_impl.bzl.

ros2/cc_defs.bzl
```diff
@@ -4,7 +4,7 @@
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
 
 C_COPTS = ["-std=c11"]
-CPP_COPTS = ["-std=c++14"]
+CPP_COPTS = ["-std=c++17"]
```
mvukov, line 7: :+1:

mvukov (commented):
Thanks again for digging into this. It's interesting that C++ part can work with humble and the rest essentially with foxy (the launch and test part).

repositories/ros2_repositories_impl.bzl should be auto-generated, please see the doc: https://github.com/mvukov/rules_ros2/blob/main/docs/ros2_repos.md. One you auto-generate the bzl file, the CI should fail, that's expected as probably the rest of the patches or the targets are obsolete. I'll help soon with fixing the rest.

examples/chatter/talker.cc
LiyouZhou: The second parameter to declare_parameter is the default value.

repositories/rcl.BUILD.bazel
LiyouZhou: Removed

LiyouZhou:
@mvukov Thank you for you review. I think I have covered all your comments. Let me know if you want anything else changed. I think the CI is not running correctly, rclpy still fails to compile in my local repo.

mvukov:
Thanks very much for your efforts. I'll look into rclpy and other upgrade to make this fully working. I'll also investigate CI issues.

mvukov:
I will continue developments in `feature/upgrade_to_humble` branch on this remote. I can confirm that there was a problem with CI. I changed the cache path and then CI fails.

mvukov:
Closing in favor of #19 .

PR #18, opened 2022-11-01
Enable shared memory backend

LiyouZhou:
Use iceoryx to enable shared memory transport for ROS2. Huge speed up for large messages such as images or point clouds.

- Add iceoryx to build system
- Integrate iceoryx with cyclonedds build
- Modified example to demonstrate shared memory transport
- Updated documentation


LiyouZhou:
@mvukov This is the follow on PR to https://github.com/mvukov/rules_ros2/pull/16, will have to retarget to main after that is merged.

WORKSPACE
```diff
@@ -25,11 +25,23 @@ pip_parse(
...
+http_archive(
+    name = "rules_foreign_cc",
+    sha256 = "6041f1374ff32ba711564374ad8e007aef77f71561a7ce784123b9b4b88614fc",
+    strip_prefix = "rules_foreign_cc-0.8.0",
+    url = "https://github.com/bazelbuild/rules_foreign_cc/archive/0.8.0.tar.gz",
+)
+
+load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
+
+rules_foreign_cc_dependencies()
```
mvukov, line 43: This should go to repositories/deps.bzl

examples/chatter/Chatter.msg
```diff
@@ -0,0 +1,3 @@
+uint64 timestamp
+char[100] data
```
mvukov, line 2: It would be nice to document this details on the fixed size.

examples/configs/BUILD.bazel
```diff
@@ -0,0 +1,4 @@
+filegroup(
+    name = "configs",
+    srcs = glob(["*"]),
+)
```
mvukov, line 4: Once there is a separate example for zero-copy stuff, I'd suggest to keep the config in there as well.

format.patch
```diff
@@ -0,0 +1,246 @@
+diff --git a/examples/actions/client.cc b/examples/actions/client.cc
```
mvukov, line 1: I don't think we need this file any more, right?

repositories/cyclonedds.BUILD.bazel
```diff
@@ -19,7 +19,7 @@ expand_template(
         "#cmakedefine DDS_HAS_SSL 1": "#undef DDS_HAS_SSL",
         "#cmakedefine DDS_HAS_TYPE_DISCOVERY 1": "#define DDS_HAS_TYPE_DISCOVERY 1",
         "#cmakedefine DDS_HAS_TOPIC_DISCOVERY 1": "#define DDS_HAS_TOPIC_DISCOVERY 1",
-        "#cmakedefine DDS_HAS_SHM 1": "#undef DDS_HAS_SHM",
+        "#cmakedefine DDS_HAS_SHM 1": "#define DDS_HAS_SHM 1",
```
mvukov, line 22: I think we should have a `select`statement to conditionally enable this feature for cyclonedds. You can use a bool flag from skylib here; [example](https://github.com/mvukov/qpoases_embedded/blob/master/BUILD.bazel#L4-L14).

repositories/cyclonedds.BUILD.bazel
```diff
@@ -251,8 +251,11 @@ cc_library(
         "src/ddsrt/include",
         "src/ddsrt/src",
     ],
-    linkopts = ["-lpthread"],
-    deps = [],
+    linkopts = [
+        "-lpthread",
+        "-lrt",
```
mvukov, line 256: Should this be a linkopt defined for iceoryx?

repositories/cyclonedds.BUILD.bazel
```diff
@@ -329,6 +332,7 @@ _DDSI_SRCS = [
     "ddsi_vnet.c",  # Manually added!
     "ddsi_wraddrset.c",  # Manually added!
     "ddsi_list_genptr.c",
+    "ddsi_shm_transport.c",
```
mvukov, line 335: How about we start using rules_foreign_cc for cyclonedds?

repositories/iceoryx.BUILD.bzl
```diff
@@ -0,0 +1,52 @@
...
+    out_static_libs = [
+        "libiceoryx_binding_c.a",
+        "libiceoryx_posh_config.a",
+        "libiceoryx_posh_roudi.a",
+        "libiceoryx_posh.a",
+        "libiceoryx_posh_gateway.a",
+        "libiceoryx_hoofs.a",
+        "libiceoryx_platform.a",
+    ],
+    out_binaries = ["iox-roudi"],
```
mvukov, line 32: Do we need this?

repositories/iceoryx.BUILD.bzl
```diff
@@ -0,0 +1,52 @@
...
+    out_include_dir = "include/iceoryx/v2.0.2/",
+    visibility = ["//visibility:public"],
+    working_directory = "iceoryx_meta",
```
mvukov, line 35: Just curious: why do we need this?

repositories/iceoryx.BUILD.bzl
```diff
@@ -0,0 +1,52 @@
...
+cc_binary(
+    name = "shared_memory_manager",
+    srcs = ["iceoryx_posh/source/roudi/application/roudi_main.cpp"],
+    deps = [":iceoryx"],
+    linkopts = [
+        "-lpthread",
+        "-lrt",
+        "-latomic",
+    ],
+    linkstatic = True,
```
mvukov, line 49: This is the default, no?

repositories/repositories.bzl
```diff
@@ -65,6 +65,33 @@ def ros2_repositories():
...
+    maybe(
+        http_archive,
+        name = "acl",
+        build_file = "@com_github_mvukov_rules_ros2//repositories:acl.BUILD.bzl",
+        sha256 = "06be9865c6f418d851ff4494e12406568353b891ffe1f596b34693c387af26c7",
+        url = "https://de.freedif.org/savannah/acl/acl-2.2.53.tar.gz",
+        strip_prefix = "acl-2.2.53",
+    )
+
+    maybe(
```
mvukov, line 86: Should be used from the ros2.repos https://github.com/ros2/ros2/blob/humble-release/ros2.repos#L46. See https://github.com/mvukov/rules_ros2/blob/main/docs/ros2_repos.md; please run the resolver once you update the yaml file.

But this should happen after the target PR is consolidated.

repositories/rmw_cyclonedds.BUILD.bazel
```diff
@@ -13,7 +13,10 @@ ros2_cpp_binary(
     ]),
     copts = ["-fvisibility=hidden"],
     includes = ["rmw_cyclonedds_cpp/include"],
-    linkopts = ["-fvisibility=hidden"],
+    linkopts = [
+        "-fvisibility=hidden",
+        "-latomic",
```
mvukov, line 18: -latomic should be defined as a linkopt of iceoryx. For the reference, this is a system library.

repositories/iceoryx.BUILD.bzl
mvukov: OK, took a look at the repo layout.

repositories/repositories.bzl
```diff
@@ -65,6 +65,33 @@ def ros2_repositories():
...
+        name = "attr",
+        build_file = "@com_github_mvukov_rules_ros2//repositories:attr.BUILD.bzl",
+        sha256 = "25772f653ac5b2e3ceeb89df50e4688891e21f723c460636548971652af0a859",
+        url = "http://download.savannah.gnu.org/releases/attr/attr-2.4.47.src.tar.gz",
+        strip_prefix = "attr-2.4.47",
+    )
+
+    maybe(
+        http_archive,
+        name = "acl",
```
mvukov, line 79: See https://github.com/eclipse-iceoryx/iceoryx/blob/6c7c3a51491082a24f383aa4d9023f737be8269b/iceoryx_platform/qnx/include/iceoryx_platform/acl.hpp#L20, iceoryx always wants to use acl from the system (rootfs). So, I would remove builds for acl and attr and just put together which apt packages need to be installed.

mvukov (commented):
OK, per https://github.com/eclipse-iceoryx/iceoryx#where-is-eclipse-iceoryx-used it looks like we can still use rmw_cyclonedds even with iceoryx.

mvukov:
A side note, please install and auto-use pre-commit ;) (`pre-commit install`). I see that you have some formatting-only commits.

repositories/repositories.bzl
mvukov: Looks like we only need `libacl1-dev`.

repositories/cyclonedds.BUILD.bazel
LiyouZhou: Whether to enable on shared memory backend is already entirely a[ run time configuration](https://github.com/mvukov/rules_ros2/pull/18/files#diff-6b53520cd82a74f6db2af93ae8e695a697faf73fc5b50f7fc43cec875cfbfc93R5). Cyclone dds is also smart enough to only use shm for publisher/subscriber pairs that fits all the requirements. It will drop back to using network in all cases. Hence I thought it would be ok to always build SHM, it certainly makes the BUILD file simpler and don't need user to do extra steps.

repositories/cyclonedds.BUILD.bazel
LiyouZhou: You mean like using a cmake rule? Could be simpler to maintain. Not sure I will have time to work on that tho.

LiyouZhou:
> OK, per https://github.com/eclipse-iceoryx/iceoryx#where-is-eclipse-iceoryx-used it looks like we can still use rmw_cyclonedds even with iceoryx.

Yeah iceoryx only serves as a backend for cyclonedds. Cyclone dds is smart enough to only use SHM for pub/sub pairs that fits all the criteria and will fall back to using network for others.

This is alway why cyclonedds with iceoryx is better than rwm_iceoryx. rwm_iceoryx will confine the ROS2 to running on only 1 system.

LiyouZhou:
>A side note, please install and auto-use pre-commit ;) (pre-commit install). I see that you have some formatting-only commits.
My local setup is a bit messed up due to some py2 py3 bs. Are those tests run as CI?

LiyouZhou:
@mvukov If you have time please take a look where I have addressed your comments.

mvukov:
>  My local setup is a bit messed up due to some py2 py3 bs. Are those tests run as CI?

Yeah, lint tests are also ran by the CI.

examples/zero_copy/listener.cc
```diff
@@ -0,0 +1,47 @@
+// Copyright 2016-2022 Open Source Robotics Foundation, Inc., Milan Vukov, wayve.ai
```
mvukov, line 1: You can remove me from this copyright. If you feel wayve should hold a copyright on this example, please add a new line -- the first line should be the same as the original.

examples/zero_copy/Chatter.msg
```diff
@@ -0,0 +1,3 @@
+uint64 timestamp
+char[4000000] data # shared memory backend can only deal with fixed size messages
```
mvukov, line 2: Would it make sense to put here 4MiB, aka 4 * 1024 * 1024?

examples/zero_copy/talker.cc
```diff
@@ -0,0 +1,59 @@
...
+      auto message = publisher_->borrow_loaned_message();
+      std::string str = "Hello, world! " + std::to_string(count_++);
+      auto copy_size = str.size() < message.get().data.size()
+                           ? str.size()
+                           : message.get().data.size();
+      std::memcpy(message.get().data.begin(), str.c_str(), copy_size);
+      message.get().data_length = copy_size;
+      message.get().timestamp =
+          std::chrono::duration_cast<std::chrono::microseconds>(
+              std::chrono::system_clock::now().time_since_epoch())
```
mvukov, line 39: Let's use std::steady_clock here.

examples/zero_copy/listener.cc
```diff
@@ -0,0 +1,47 @@
...
+  MinimalSubscriber() : Node("minimal_subscriber") {
+    subscription_ = create_subscription<chatter_interface::msg::Chatter>(
+        "topic", 10, [this](chatter_interface::msg::Chatter::SharedPtr msg) {
+          auto current_timestamp =
+              std::chrono::duration_cast<std::chrono::microseconds>(
+                  std::chrono::system_clock::now().time_since_epoch())
+                  .count();
+          std::string str{msg->data.begin(),
+                          msg->data.begin() + msg->data_length};
+          RCLCPP_INFO(get_logger(), "I heard: '%s', delay %lu us", str.c_str(),
```
mvukov, line 32: I'd print the delay first. The max size is 4MB in size, so, if someone modifies the example to really utilize the size this won't be readable.

repositories/iceoryx.BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+    out_data_dirs = ["lib/cmake"],
+    visibility = ["//visibility:public"],
+    linkopts = [
+        "-lpthread",
+        "-lrt",
+        "-latomic",
+        "-lacl",
+    ],
+    working_directory = "iceoryx_meta",
+    tags = ["requires-network"],
```
mvukov, line 42: Interesting. Do you have an idea what they pull from the network?

repositories/iceoryx.BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+        "libiceoryx_platform.a",
+    ],
+    out_include_dir = "include/iceoryx/v2.0.2/",
+    out_data_dirs = ["lib/cmake"],
+    visibility = ["//visibility:public"],
+    linkopts = [
+        "-lpthread",
+        "-lrt",
+        "-latomic",
+        "-lacl",
```
mvukov, line 39: Please alph-sort those opts. Also, please add a line in the readme of the example that people need to install libacl1-dev. AFAIK, libatomic is a system lib.

repositories/iceoryx.BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+    },
+    lib_source = ":all_srcs",
+    out_static_libs = [
+        "libiceoryx_binding_c.a",
+        "libiceoryx_posh_config.a",
+        "libiceoryx_posh_roudi.a",
+        "libiceoryx_posh.a",
+        "libiceoryx_posh_gateway.a",
+        "libiceoryx_hoofs.a",
+        "libiceoryx_platform.a",
```
mvukov, line 30: Please alph-sort the libs.

repositories/iceoryx.BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+    name = "iceoryx",
+    build_args = [
+        "--",  # <- pass options to the native tool
+        "-j4",
+    ],
+    cache_entries = {
+        "BUILD_SHARED_LIBS": "OFF",
+        "CCACHE": "OFF",
+        "CMAKE_BUILD_TYPE": "Release",
+        "CMAKE_CXX_FLAGS": "-I$EXT_BUILD_DEPS/acl/include -L$EXT_BUILD_DEPS/acl/lib",
```
mvukov, line 20: This is not needed any more.

examples/zero_copy/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+```sh
+export CYCLONEDDS_URI=file://<path to rules_ros2>/examples/zero_copy/configs/cyclonedds.xml
+bazel run //examples/chatter:talker
+```
+
+In another terminal run
+
+```sh
+bazel ruexport CYCLONEDDS_URI=file://<path to rules_ros2>/examples/zero_copy/configs/cyclonedds.xml
```
mvukov, line 19: Nit: export CYCLONEDDS_URI.

examples/zero_copy/Chatter.msg
```diff
@@ -0,0 +1,3 @@
...
+uint64 data_length
```
mvukov, line 3: I would spice this up by adding `count` field such that the listener can very that there are no dropped msgs. WDYT?

examples/zero_copy/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+```sh
+bazel run @iceoryx//:shared_memory_manager
+```
+
+Then, for each node started, specify the CycloneDDS config file:
+
+```sh
+export CYCLONEDDS_URI=file://<path to rules_ros2>/examples/zero_copy/configs/cyclonedds.xml
+bazel run //examples/chatter:talker
```
mvukov, line 13: Off the top of my head, I would suggest interested people run this in opt mode, as e.g. `--config=opt`. WDYT?

In principle, as iceoryx and cyclonedds are compiled in release mode, that should already help.

README.md
```diff
@@ -22,6 +22,7 @@ At the moment it is supported:
 - Defining ROS2 deployments with `ros2_launch` Bazel macro.
 - Defining ROS2 tests with `ros2_test` Bazel macro.
 - Only CycloneDDS middleware can be interfaced at the moment.
+- Zero copy transport via shared memory backend for CycloneDDS.
```
mvukov, line 25: Perhaps mention that we use iceoryx here.

examples/zero_copy/talker.cc
```diff
@@ -0,0 +1,59 @@
...
+#include <cstring>
+#include <memory>
+
+#include "chatter_interface/msg/chatter.hpp"
+#include "rclcpp/rclcpp.hpp"
+
+class MinimalPublisher : public rclcpp::Node {
+ public:
+  MinimalPublisher() : Node("minimal_publisher"), count_(0) {
+    declare_parameter("callback_period_ms", 500);
```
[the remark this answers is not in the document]
LiyouZhou: Tried that, makes too much console print to read. Are you ok to leave as is?

examples/zero_copy/Chatter.msg
LiyouZhou: 4MiB exceeded their default reserved shared memory pool size, I used 3 MiB instead.

repositories/iceoryx.BUILD.bazel
LiyouZhou: the toml parsing library i think.

repositories/iceoryx.BUILD.bazel
LiyouZhou: Run into issue with this actually.

The order of `out_static_libs` in the cmake rule is the order being passed to the linker in the final cc_binary rule and this order matters. The rule is the library that needs symbols must be first, then the library that resolves the symbol after. Hence this list can only be in dependency order.

LiyouZhou:
@mvukov addressed you comments. Would you mind taking another look? Thx.

repositories/cyclonedds.BUILD.bazel
mvukov: Already in the main branch.

mvukov:
Many thanks for the efforts. I will finish up some things in [feature/enable_shared_memory_backend](https://github.com/mvukov/rules_ros2/tree/feature/enable_shared_memory_backend) branch. E.g. I want to make iceoryx conditional.

mvukov:
Closing in favor of #23 .

PR #20, opened 2022-11-07
Allow overriding 'strip_prefix' in resolver

LiyouZhou:


mvukov (approved):

repositories/private/resolver.py
```diff
@@ -180,7 +180,8 @@ def main():
     for name, sha_sum in names_to_sha_sums.items():
         repo_mappings[name]['sha256'] = sha_sum
     for name, version in names_to_versions.items():
-        repo_mappings[name]['strip_prefix'] = f'{name}-{version}'
+        if 'strip_prefix' not in repo_mappings[name].keys():
```
mvukov, line 183: btw, this will also work if you drop `.keys()`.

PR #21, opened 2022-11-08
Add visualization_msgs and fix sensor_msgs includes

ahans:
I can split this into two separate PRs if you prefer (or at least two commits in a single PR). Also, I started out with extending `ros2_interface_library` to support a `hdrs` attributes, but then realized that the not-yet-included header files are actually only useful for C++, so having the private `cpp_ros2_interface_library` and wrapping this with a `ros2_cpp_library` that includes the headers seemed like a good solution. Please let me know if you'd prefer a different approach.


mvukov (approved):
LGTM! The solution is just fine. Thanks for you effort! :rocket:

PR #29, opened 2022-11-24
Add Python targets for common_interfaces

ahans:


repositories/common_interfaces.BUILD.bazel
```diff
@@ -92,6 +104,22 @@ ros2_cpp_library(
     deps = [":_cpp_sensor_msgs"],
 )
 
+py_ros2_interface_library(
+    name = "py_sensor_msgs",
```
mvukov, line 108: Do you think it makes sense to do here something similar you made for the C++ version? In particular, make this one private and name it _py_sensor_msgs and the lib below name as py_sensor_msgs.

ahans: I can sure do that, but that wouldn't exactly reflect how packages are organized in ROS. There, you have a single package `sensor_msgs` containing the message definitions and the C++ support code. The Python code is in a separate package `sensor_msgs_py`.  However, I didn´t want to have a package `py_sensor_msgs` for the Python message code and `sensor_msgs_py` with the support code. That's why I renamed it. However, I guess now that we deviate from the ROS2 package structure anyway, we can follow the same pattern as we have for C++.

mvukov (approved):
LGTM! :rocket:

PR #37, opened 2022-12-20
Feature/add image common

dbruggner:
* Cherry-pick for support of multiple base classes in pluginlib
* Extent ros2_launch to set AMENT_PREFIX_PATH
* Add image_common package

### DIFF
--- a/repositories/image_common.BUILD.bazel
+++ b/repositories/image_common.BUILD.bazel
@@ -34,61 +34,67 @@ ros2_cpp_library(
 )
 
 ros2_cpp_library(
-    name = "image_transport_common_hdrs",
-    hdrs = glob([
+    name = "image_transport_common",
+    srcs = [
+        "image_transport/src/camera_common.cpp",
+        "image_transport/src/publisher.cpp",
+        "image_transport/src/single_subscriber_publisher.cpp",
+    ],
+    hdrs = [
         "image_transport/include/image_transport/camera_common.h",
         "image_transport/include/image_transport/camera_common.hpp",
-        "image_transport/include/image_transport/camera_publisher.h",
-        "image_transport/include/image_transport/camera_publisher.hpp",
-        "image_transport/include/image_transport/camera_subscriber.h",
-        "image_transport/include/image_transport/camera_subscriber.hpp",
         "image_transport/include/image_transport/exception.h",
         "image_transport/include/image_transport/exception.hpp",
-        "image_transport/include/image_transport/image_transport.h",
-        "image_transport/include/image_transport/image_transport.hpp",
         "image_transport/include/image_transport/loader_fwds.h",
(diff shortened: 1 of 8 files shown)


mvukov:
Thanks for the contribution. I ironed out ros plugin stuff, please rebase image-common stuff on the main branch -- fine with me if you want to do this in a clean new branch.

dbruggner:
@mvukov Done.

repositories/image_common.BUILD.bazel
```diff
@@ -0,0 +1,147 @@
...
+    deps = [
+        "@ros2_common_interfaces//:cpp_sensor_msgs",
+        "@ros2_rclcpp//:rclcpp",
+        "@ros2_rcpputils//:rcpputils",
+        "@yaml-cpp",
+    ],
+)
+
+ros2_cpp_library(
+    name = "image_transport",
```
mvukov, line 37: Didn't look into the original code thoroughly, but, I would expect that the plugin could be data dep of this target (that way the plugin also doesn't need the public visibility). Please check if this is possible.

Furthermore, I would expect a structure as following:
```Bazel

cc_library(
  name = "image_transport_common_hdrs",
  hdrs = <explicitly defined>,
  ...
)

ros2_cpp_library(
  name = "image_transport",
  deps = [":image_transport_common_hdrs"],
  data = [":raw_pub_sub],
  hdrs = [],  # explicitly defined such that deps are clear,
  # should not include common (as those come in via deps), raw_publisher, raw_subscriber hdrs.
  ...
)

cc_library(
  name = "raw_publisher",
  deps = [":image_transport_common_hdrs"],
  ...
)

cc_library(
  name = "raw_subscriber",
  deps = [":image_transport_common_hdrs"],
  ...
)

ros2_plugin(
  name = "raw_pub_sub_plugin",
  srcs = [...],
  deps = [":raw_publisher", "raw_subscriber"],
)

```

mvukov: On a second thought, image_transport should depend on raw_publisher and raw_subscriber as well -- but only on the interfaces, not on any code related to the concrete impl of plugins.

dbruggner: I updated the code.

mvukov:
I updated the code a bit and added some simple tests.

yokyj:
Hi, I would like to ask why we used foreign_cc cmake for yaml-cpp here while the 0.7 branch of the repo supports bazel already.
BTW, I have encountered a problem that cmake for yaml-cpp failed.
```
bazel build --verbose_failures --jobs=3 --config=gen_sym   @com_github_mvukov_rules_ros2//ros2/test/rosbag:recorder


output 'external/yaml_cpp/yaml_cpp/lib/libyaml-cpp.a' was not created
```

PR #43, opened 2023-01-15
Add QT and turtlesim example

AndreasKnoblach:
Adds QT support and turtlesim example.

The very vast majority of changes is from copying the original turtlesim example (Commit: [Add turtlesims as is from tutorial to examples](https://github.com/mvukov/rules_ros2/pull/43/commits/2a8f20b14a13c146ba82dc072d8e1b65243d78e1)).


examples/turtlesim/src/turtle_frame.cpp
```diff
@@ -0,0 +1,304 @@
...
+  turtles.append("humble.png");
+  turtles.append("rolling.png");
+
+  /*
+   * QString images_path = (ament_index_cpp::get_package_share_directory("turtlesim") + "/images/").c_str();
+   * AndreasKnoblach 2023-01-15: get_package_share_directory cause crash because
+   *   Environment variable 'AMENT_PREFIX_PATH' is not set or empty
+   * Since this is only needed to determine the images_path, it is replaced by hard coding the path.
+   */
+  // QString images_path{"<complete path to rules_ros2>/examples/turtlesim/images/"}; //inentionally comment to trigger compile error
```
AndreasKnoblach, line 97: ATM, `ament_index_cpp::get_package_share_director` is not yet working properly. Hence the path to the turtle images needs to be configured manually, here.

examples/turtlesim/BUILD.bazel
```diff
@@ -0,0 +1,87 @@
...
+        ":cpp_turtlesim_interface",
+        "@qt//:qt_core",
+        "@qt//:qt_gui",
+        "@ros2_common_interfaces//:cpp_geometry_msgs",
+        "@ros2_rclcpp//:rclcpp",
+        "@ros2_rclcpp//:rclcpp_action",
+    ],
+)
+
+ros_qt_cc_library(
```
AndreasKnoblach, line 31: Once the `ament_index_cpp::get_package_share_director` issues is fixed, the images are to be added here, too.

examples/third_party/qt/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
+cc_library(
```
AndreasKnoblach, line 1: Maybe this files and qt.bzl can be moved out of the example folder. Did not get working in a first shot.

AndreasKnoblach:
Thx for your comment Milan.

I agree to your concerns and have actually expected this. My reason to use QT anyhow was because QT is widely used in many ROS examples and I considered an example how to integrated it as helpful. While I'd be interested in porting it to IMGUI, I do not need it for my self and do not see that I have the cycles for this, ATM. Sorry.

Feel free to close this PR
Andreas

mvukov:
Hi Andreas, I fully understand you. Let's put the PR in the fridge at the moment, and maybe we or someone else will have time to take a look at this in the future. Thanks once again for the initial push!

AndreasKnoblach:
Thx for your understanding!

BTW, which ImGui backend would you use?

mvukov:
Hi, I already experimented with DearImGui with OpenGL+glfw backend on Linux, and this is quite easy to set up on Linux + Bazel; and it works like a charm. So, I think I would start with that. Any other ideas?

PR #45, opened 2023-01-19
Fix flaky image_transport test

ahans:
This test failed when the GTest binary fininished before we enter `TestPlugins.test_plugins()`. Adding the `keep_alive` decorator fixes that.


mvukov:
Very nice, thanks. I also fixed `ros2/test/pluginlib/tests.py` for the same reason -- I saw it sometimes be flaky as well.

PR #48, opened 2023-01-25
Add static transform publisher node

dbruggner:


mvukov:
Many thanks for the contribution!

PR #52, opened 2023-02-04
Enable building spdlog for QNX

matzipan:
In QNX pthread is bundled with the rest of the libc


mvukov (approved):
Many thanks for your contribution! :rocket:

PR #54, opened 2023-02-04
Enable building cyclonedds for QNX

matzipan:
In QNX pthread is bundled with libc.

My full patch actually looks like this:
```patch
diff --git a/BUILD.bazel b/BUILD.bazel
old mode 100755
new mode 100644
index 1d33e9b..f0047f9
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -24,6 +24,7 @@ filegroup(
 
 cache_entries = {
     "CMAKE_BUILD_TYPE": "Release",
+    "CMAKE_SYSTEM_NAME": "Foo",
     "CMAKE_POSITION_INDEPENDENT_CODE": "ON",  # Must be set!
     "BUILD_SHARED_LIBS": "OFF",
     # CycloneDDS specific options.
@@ -38,7 +39,8 @@ cache_entries = {
     "ENABLE_LIFESPAN": "ON",
     "ENABLE_NETWORK_PARTITIONS": "ON",
     "ENABLE_SECURITY": "OFF",
-    "ENABLE_SOURCE_SPECIFIC_MULTICAST": "ON",
+    "ENABLE_SOURCE_SPECIFIC_MULTICAST": "OFF",
+    "ENABLE_IPV6": "OFF",
     "ENABLE_SSL": "OFF",  # TODO(mvukov) Here we could use openssl/boringssl.
     "ENABLE_TOPIC_DISCOVERY": "OFF",
     "ENABLE_TYPE_DISCOVERY": "OFF",
@@ -70,7 +72,14 @@ cmake(
     }),
     generate_args = ["-GNinja"],
     lib_source = ":all_srcs",
-    linkopts = ["-lpthread"],
+    linkopts = select(
+        {
+            "@platforms//os:linux": ["-lpthread"],
+            "@platforms//os:macos": ["-lpthread"],
+            "@platforms//os:qnx": [],
+        },
+        no_match_error = "Only Linux, macOS and QNX are supported!",
+    ),
     out_static_libs = ["libddsc.a"],
     visibility = ["//visibility:public"],
     deps = select({

```

But I'm not sure if this is acceptable upstream.

### DIFF
--- a/repositories/cyclonedds.BUILD.bazel
+++ b/repositories/cyclonedds.BUILD.bazel
@@ -70,7 +70,14 @@ cmake(
     }),
     generate_args = ["-GNinja"],
     lib_source = ":all_srcs",
-    linkopts = ["-lpthread"],
+    linkopts = select(
+        {
+            "@platforms//os:linux": ["-lpthread"],
+            "@platforms//os:macos": ["-lpthread"],
+            "@platforms//os:qnx": [],
+        },
+        no_match_error = "Only Linux, macOS and QNX are supported!",
+    ),
     out_static_libs = ["libddsc.a"],
     visibility = ["//visibility:public"],
     deps = select({
--- a/repositories/cyclonedds.BUILD.bazel
+++ b/repositories/cyclonedds.BUILD.bazel
@@ -12,13 +12,13 @@ bool_flag(
 )
 
 config_setting(
-    name = "with_shm",
+    name = "enable_shm_on",
     flag_values = {":enable_shm": "True"},
     visibility = ["//visibility:public"],
 )
 
 config_setting(
-    name = "without_shm",
+    name = "enable_shm_off",
     flag_values = {":enable_shm": "False"},
     visibility = ["//visibility:public"],
 )
@@ -35,7 +35,7 @@ selects.config_setting_group(
     name = "linux_or_macos_with_shm",
     match_all = [
         ":linux_or_macos",
-        ":with_shm",
+        ":enable_shm_on",
     ]
 )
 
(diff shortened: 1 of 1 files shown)


mvukov (approved):
:rocket:

mvukov:
Thanks for the contribution.

You can see that I already split common settings and (non-)iceoryx specific options. Off the topic of my head, I think this should help here: https://github.com/bazelbuild/bazel-skylib/blob/main/docs/selects_doc.md#selectsconfig_setting_group, such that you have at least:

non-iceoryx+linux_or_macos
non-iceoryx+qnx
iceoryx+linux_or_macos

if you are not interested in iceoryx+qnx that's ok, but please leave a comment in your PR. For concatenating dicts I would still try to use dicts.add.

mvukov:
So, maybe obvious, but I'll leave note here: I suggest that should should further split cache_entries: keep common stuff in there and then create linux_or_macos specific settings and qnx specific settings.

If linux_or_macos constraint is cumbersome to set up, just leave out macos -- we don't have macos CI and I haven't tested this repo on macos since a long while.

matzipan:
I absolutely Bazel's stubornness around nested selects. It makes a lot of things so much messier.

I'll see if I can find a nice solution.

mvukov:
Lemme know if you need help. BTW, it would be nice if you could eventually write some docs about qnx cross-compilation.

matzipan:
> BTW, it would be nice if you could eventually write some docs about qnx cross-compilation

Nothing special. But I will trigger a discussion about open sourcing the QNX toochain files that I wrote ages ago.

PR #60, opened 2023-02-17
Add service and param commands

dbruggner:


ros2/param.bzl
```diff
@@ -0,0 +1,25 @@
+""" Implements a macro for setting up a ROS2 param app.
```
mvukov, line 1: What's the motivation for ros2_param macro? I though ros2_param binary (defined in https://github.com/mvukov/rules_ros2/blob/main/ros2/BUILD.bazel#L32) is sufficient. BTW, you can alias it in your workspace as https://github.com/mvukov/rules_ros2/blob/main/examples/BUILD.bazel#L2.

ros2/ros2_service.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022 Milan Vukov
```
mvukov, line 1: It's your work, just put your name here. And update year to 2023 ;).

mvukov (commented):
Nice work :)

mvukov:
BTW, please update the main readme. Please add ros2_service in the list of utils.

ros2/param.bzl
dbruggner: you're right!

dbruggner:
Done

mvukov (approved):
:rocket:

PR #63, opened 2023-02-28
Add rosbag mcap plugin

dbruggner:


dbruggner:
The mcap plugin cannot find the message definitions (`.msg`) files that's why data is recorded but no message schema is stored in the resulting mcap file.

dbruggner:
It's working now.

.bazelrc
```diff
@@ -24,3 +24,6 @@ test --build_tests_only
 # as the user configuration should be able to overwrite flags from this file.
 # See https://docs.bazel.build/versions/master/best-practices.html#bazelrc
 try-import %workspace%/user.bazelrc
+
+# C++17 is required for the MCAP support in ROS Bag
+build --cxxopt='-std=c++17'
```
mvukov, line 29: Please import https://github.com/mvukov/rules_ros2/blob/main/ros2/cc_defs.bzl#L7 and use where needed instead.

examples/.bazelrc
```diff
@@ -24,3 +24,6 @@ test --build_tests_only
 # as the user configuration should be able to overwrite flags from this file.
 # See https://docs.bazel.build/versions/master/best-practices.html#bazelrc
 try-import %workspace%/user.bazelrc
+
+# C++17 is required for the MCAP support in ROS Bag
+build --cxxopt='-std=c++17'
```
mvukov, line 29: Same here.

repositories/lz4.BUILD.bazel
```diff
@@ -0,0 +1,62 @@
...
+    ],
+    strip_include_prefix = "lib/",
+    visibility = ["//visibility:private"],
+)
+
+cc_library(
+    name = "lz4_hc",
+    srcs = [
+        "lib/lz4hc.c",
+        #        "lib/lz4opt.h",
```
mvukov, line 34: rm dead code, here and elsewhere.

repositories/lz4.BUILD.bazel
```diff
@@ -0,0 +1,62 @@
...
+)
+
+# lz4_hc includes a .c file...
+cc_library(
+    name = "lz4_lz4c_include",
+    hdrs = [
+        "lib/lz4.c",
+    ],
+    strip_include_prefix = "lib/",
+    visibility = ["//visibility:private"],
```
mvukov, line 27: the default visibility is private, this is not needed.

repositories/lz4.BUILD.bazel
mvukov: btw, strange that the file isn't used.

repositories/repositories.bzl
```diff
@@ -124,6 +124,24 @@ def ros2_repositories():
         urls = ["https://github.com/facebook/zstd/releases/download/v1.5.2/zstd-1.5.2.tar.gz"],
     )
 
+    maybe(
+        http_archive,
+        name = "lz4",
+        build_file = "@com_github_mvukov_rules_ros2//repositories:lz4.BUILD.bazel",
+        sha256 = "4ec935d99aa4950eadfefbd49c9fad863185ac24c32001162c44a683ef61b580",
+        strip_prefix = "lz4-1.9.3",
```
mvukov, line 132: Is this version referenced by mcap version below?

repositories/rosbag2.BUILD.bazel
```diff
@@ -12,6 +12,7 @@ load(
     "ros2_interface_library",
 )
 load("@com_github_mvukov_rules_ros2//ros2:plugin.bzl", "ros2_plugin")
+load("@rules_cc//cc:defs.bzl", "cc_library")
```
mvukov, line 15: Please use macros from //ros2:cc_defs.bzl if you need them.

repositories/rosbag2.BUILD.bazel
```diff
@@ -32,28 +33,78 @@ ros2_cpp_library(
     ],
 )
 
-ros2_plugin(
+cc_library(
```
mvukov, line 36: This can be reverted AFAICS.

repositories/rosbag2.BUILD.bazel
```diff
@@ -32,28 +33,78 @@ ros2_cpp_library(
...
+        "@ros2_pluginlib//:pluginlib",
+        "@ros2_rcpputils//:rcpputils",
+        "@ros2_rcutils//:rcutils",
+        "@sqlite",
+        "@yaml_cpp",
+    ],
+    alwayslink = True,
+)
+
+cc_library(
```
mvukov, line 55: ros2_cpp_library

repositories/rosbag2.BUILD.bazel
```diff
@@ -32,28 +33,78 @@ ros2_cpp_library(
...
+    name = "mcap_vendor",
+    srcs = ["mcap_vendor/src/main.cpp"],
+    deps = [
+        "@mcap",
+        "@zstd",
+    ],
+)
+
+cc_library(
```
mvukov, line 64: ros2_plugin

repositories/rosbag2.BUILD.bazel
```diff
@@ -32,28 +33,78 @@ ros2_cpp_library(
...
+        ":mcap_vendor",
+        ":rosbag2_storage",
+        "@ros2_pluginlib//:pluginlib",
+        "@ros2_rcpputils//:rcpputils",
+        "@ros2_rcutils//:rcutils",
+    ],
+    alwayslink = True,
+)
+
+ros2_plugin(
```
mvukov, line 91: OK, I see what you want to achieve. Having pattern where cc_library the only has deps should be avoided IMO. Mainly because people tend to add deps and forget to remove them. We'll document what plugins are available and how to include extra plugins. We can always include the default one, and users will know how to include additional ones if they want.

repositories/rosbag2.BUILD.bazel
```diff
@@ -81,7 +132,9 @@ ros2_cpp_library(
         "rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp",
     ],
     hdrs = glob(["rosbag2_cpp/include/**/*.hpp"]),
-    data = [":rosbag2_storage_default_plugins"],
+    data = [
+        ":rosbag2_storage_plugins",
```
mvukov, line 136: Or we can opt here to put here more plugins.

ros2/ament.bzl
```diff
@@ -147,6 +147,18 @@ def _ros2_ament_setup_rule_impl(ctx):
         )
         outputs.append(dynamic_library)
 
+    for dep in ctx.attr.idl_deps:
```
mvukov, line 150: Calling .to_list is inefficient. BTW, I think you'll need:

```
    idls = depset(
        transitive = [
            depset([
                struct(
                    package_name = dep.label.name,
                    srcs = dep[Ros2InterfaceInfo].info.srcs
                )
            ])
            for dep in ctx.attr.idl_deps
        ],
    ).to_list()
```
and then you that struct in the loop.

Otherwise, nice work! :+1:

ros2/ament.bzl
```diff
@@ -147,6 +147,18 @@ def _ros2_ament_setup_rule_impl(ctx):
...
+        package_name = dep[Ros2IdlPluginAspectInfo].plugins.to_list()[-1].package_name
+        for src in dep[Ros2InterfaceInfo].info.srcs:
+            src_file = ctx.actions.declare_file(
+                paths.join(prefix_path, "share", package_name, "msg", src.basename),
```
mvukov, line 154: I would also declare ament package xmls if not already defined (take a look above a bit).

repositories/rosbag2.BUILD.bazel
dbruggner: The problem was, that if I add both plugins separately here, two plugins.xml files are being written which conflict.

repositories/repositories.bzl
dbruggner: Yes, mcap references this specific version.

dbruggner:
Addressed your comments, only one regarding the handling of multiple plugins is left, added a comment.

repositories/rosbag2.BUILD.bazel
mvukov: I fixed this issue in #69 , you can split the plugins now.

ros2/ament.bzl
mvukov: Another thing: `idl.srcs` has all kinds of IDL files (actions, msgs, srvs), so I'd propose you filter here `.msg` files here.

dbruggner:
@mvukov it's ready now.

Issue #57, opened 2023-02-07
RMW_LIBRARY_PATH patch creates issues when running without Bazel


matzipan:
The `rmw_implementation_library_path.patch` creates a nice experience when running via bazel, but it creates a lot of unecessary annoyance when running without. For example, it causes the node processes to look for the `rmw` lib in `external/ros2_rmw_cyclonedds/librmw_cyclonedds.so`.

I'm looking to improve this system a bit. Which solution would be acceptable for you? 

What if I change the patch to first check at `RMW_LIBRARY_PATH` and if it's not there fall back to the normal code?

matzipan:
Maybe something like:

```
std::shared_ptr<rcpputils::SharedLibrary>
load_library()
{
  // The logic to pick the RMW library to load goes as follows:
  //
  // 1. If the user specified the library to use via the RMW_IMPLEMENTATION
  //    environment variable, try to load only that library.
  // 2. Otherwise, try to load the default RMW implementation.
  // 3. If that fails, try loading all other implementations available in turn
  //    until one succeeds or we run out of options.

  const std::string library_path = RMW_LIBRARY_PATH;
  try {
    return std::make_shared<rcpputils::SharedLibrary>(library_path.c_str());
  } catch (const std::exception & e) {}

  std::string env_var;
  try {
    env_var = rcpputils::get_env_var("RMW_IMPLEMENTATION");
  } catch (const std::exception & e) {
    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
      "failed to fetch RMW_IMPLEMENTATION "
      "from environment due to %s", e.what());
    return nullptr;
  }

  // User specified an RMW, attempt to load that one and only that one
  if (!env_var.empty()) {
    return attempt_to_load_one_rmw(env_var);
  }

  // User didn't specify, so next try to load the default RMW
  std::shared_ptr<rcpputils::SharedLibrary> ret;

  ret = attempt_to_load_one_rmw(STRINGIFY(DEFAULT_RMW_IMPLEMENTATION));
  if (ret != nullptr) {
    return ret;
  }

  // OK, we failed to load the default RMW.  Fetch all of the ones we can
  // find and attempt to load them one-by-one.
  rmw_reset_error();
  const std::map<std::string, std::string> packages_with_prefixes = ament_index_cpp::get_resources(
    "rmw_typesupport");
  for (const auto & package_prefix_pair : packages_with_prefixes) {
[...]

PR #65, opened 2023-02-28
Do not rely on `external` symlink for RMW_LIBRARY_PATH

### LINKED ISSUE #64: Using `--nolegacy_external_runfiles` will break loading `librmw_cyclonedds.so`
If I add `--nolegacy_external_runfiles` to my .bazelrc as recommended in https://docs.aspect.build/guides/bazelrc/#performance-options (see below)
```sh
# Do not build runfiles symlink forests for external repositories under
# `.runfiles/wsname/external/repo` (in addition to `.runfiles/repo`). This reduces runfiles &
# sandbox creation times & prevents accidentally depending on this feature which may flip to off by
# default in the future. Note, some rules may fail under this flag, please file issues with the rule
# author.
# Docs: https://bazel.build/reference/command-line-reference#flag--legacy_external_runfiles
build --nolegacy_external_runfiles
run --nolegacy_external_runfiles
test --nolegacy_external_runfiles
```

Then code depending on rclcpp will fail with
```
C++ exception with description "failed to initialize rcl init options: failed to load shared library 'external/ros2_rmw_cyclonedds/librmw_cyclonedds.so' due to dlopen error: external/ros2_rmw_cyclonedds/librmw_cyclonedds.so: cannot open shared object file: No such file or directory, at external/ros2_rcutils/src/shared_library.c:99, at external/ros2_rmw_implementation/rmw_implementation/src/functions.cpp:90, at external/ros2_rcl/rcl/src/rcl/init_options.c:75" thrown in the test body.
```

This seems very related to https://github.com/mvukov/rules_ros2/issues/57

ahans:
We use a different relative path to librmw_cyclonedds.so to be compatible with Bazel's `--nolegacy_external_runfiles` option. Since we depend on the respective target, it can still be found in the runfiles tree, only in a different place.

Fixes #64

### DIFF
--- a/repositories/rmw_implementation.BUILD.bazel
+++ b/repositories/rmw_implementation.BUILD.bazel
@@ -16,7 +16,7 @@ ros2_cpp_library(
     ],
     includes = ["include"],
     local_defines = [
-        "RMW_LIBRARY_PATH=\\\"external/ros2_rmw_cyclonedds/librmw_cyclonedds.so\\\"",
+        "RMW_LIBRARY_PATH=\\\"../ros2_rmw_cyclonedds/librmw_cyclonedds.so\\\"",
     ],
     visibility = ["//visibility:public"],
     deps = [


mvukov:
This is a good solution for the moment, thanks. I hoped we could use https://github.com/bazelbuild/bazel/blob/master/tools/cpp/runfiles/runfiles_src.h, but looks like it's difficult to be used. Plus `BAZEL_CURRENT_REPOSITORY` is empty on my system -- strange; dunno if that would help at all. We can revisit this later. FWIW, it's very easy to use runfiles library from rules_python...

ahans:
Using the runfiles library would make us more robust against changes to the layout of the runfiles tree. It's possible it would also enable some support for Linux and macOS. But it would not fix the issue of #57 when you want to use a binary directly and not as part of `bazel run`.

I wonder how you would want to integrate the runfiles library here. Change the patch that currently injects `RMW_LIBRARY_PATH` to use the runfiles library? Or have a wrapper that uses the runfiles library to set an environment variable that you than use?

We would probably need a combination of `$(rlocationpath @ros2_rmw_cyclonedds//:rmw_cyclonedds)` in a `BUILD.bazel` file (probably the one I changed in this PR) and then call `runfiles->Rlocation(RMW_LIBRARY_PATH);` from C++ to fully resolve the path. But it would still be a path pointing to some runfiles tree, so we would still rely on that. Maybe it would create an absolute path, so that we could at least call a node binary from anywhere directly.

Btw, `BAZEL_CURRENT_REPOSITORY` is not an environment variable. It's injected as a pre-processor define via `-DBAZEL_CURRENT_REPOSITORY=...`. For me it is set, so things should work as expected.

ahans:
I quickly tried it out.

This goes into a `BUILD.bazel`:
```
cc_binary(
    name = "hello",
    srcs = ["hello.cc"],
    data = [
        "@ros2_rmw_cyclonedds//:rmw_cyclonedds",
    ],
    local_defines = [
        "RMW_LIBRARY_PATH=\\\"$(rlocationpath @ros2_rmw_cyclonedds//:rmw_cyclonedds)\\\"",
    ],
    deps = [
        "@bazel_tools//tools/cpp/runfiles",
    ],
)
```

And here's `hello.cc`:
```
#include "tools/cpp/runfiles/runfiles.h"

#include <iostream>

int main(int argc, char** argv) {
  auto runfiles = bazel::tools ::cpp::runfiles::Runfiles::Create(
      argv[0], BAZEL_CURRENT_REPOSITORY);
  std::cout << "RMW_LIBRARY_PATH=" << RMW_LIBRARY_PATH << std::endl;
  std::cout << runfiles->Rlocation(RMW_LIBRARY_PATH) << std::endl;
}
```

Output is this:
```
❯ bazel run ros2:hello
INFO: Analyzed target //ros2:hello (1 packages loaded, 2 targets configured).
INFO: Found 1 target...
Target //ros2:hello up-to-date:
  bazel-bin/ros2/hello
INFO: Elapsed time: 0.249s, Critical Path: 0.17s
INFO: 3 processes: 1 internal, 2 linux-sandbox.
INFO: Build completed successfully, 3 total actions
INFO: Running command line: bazel-bin/ros2/hello
RMW_LIBRARY_PATH=ros2_rmw_cyclonedds/librmw_cyclonedds.so
/home/ahans/.cache/bazel/_bazel_ahans/c02197da98e070ec8a47cb18b5c98285/execroot/com_github_mvukov_rules_ros2/bazel-out/k8-fastbuild/bin/external/ros2_rmw_cyclonedds/librmw_cyclonedds.so
```

So it gives us an absolute path. Not too bad. Not sure it would solve @matzipan's use-case from #57.

ahans:
Btw, in real code I would wrap the raw pointer in a `std::unique_ptr`, so we don't accidentally leak the `runfiles` object. I wonder why they didn't do that in the first place and return a raw pointer instead. Or, even simpler, just return the `Runfiles` instance by value?!

ahans:
And another btw, I found this talk useful: https://www.youtube.com/watch?v=5NbgUMH1OGo

mvukov:
Thanks for the link! :) I fixed the issue in a more general way in #66 . `bazel::tools ::cpp::runfiles::Runfiles::Create` didn't work out for me without `argv[0]` set to `{}` -- which is the case I'd need it in a rclcpp patch. Luckily, the patch doesn't need to be modified.

PR #79, opened 2023-03-08
Use full repo name everywhere

ahans:


mvukov:
For private stuff and tests I don't think we need this AFAIK. Have you faced any problems with this?

mvukov:
But, OK, being consistent won't hurt :)

ahans:
No problems, I just noticed it in one of the tests that I wanted to use as a blueprint for something else and couldn't just copy the `load` statement. ;-) But yes, main motivation would be about being consistent.

PR #91, opened 2023-03-23
Add ros2_{cc,py}_binary macros

ahans:
I ended up using the straightforward approach without unifying the `sh_launcher` instantiations. I wasn't able to come up with a nice way that would reduce the redundancies in the calls and at the same time keep the flexibility and feel right. One issue I had was that `sh_launcher` creates the setup without being an executable. I will add some more comments next to the code.


ros2/py_defs.bzl
```diff
@@ -3,48 +3,68 @@
...
         name = name,
         srcs = [launcher],
-        data = [test_target_symlink],
-        tags = tags,
+        data = [target_impl_symlink],
+        tags = kwargs.pop("tags", []),
+        visibility = kwargs.pop("visibility", None),
     )
+
+def ros2_py_binary(name, srcs, main, set_up_ament = False, **kwargs):
```
thomasegriffith, line 48: For my own understanding: under what circumstances should `set_up_ament` be `True`?  Is it only when `"@ros2_rosbag2//:rosbag2_py"` is a dependency?

ahans: You will need to set `set_up_ament` to be `True` whenever the plugin mechanism is involved. Rosbag support is one example, there we need to load the storage plugins. Another example is the [test](https://github.com/mvukov/rules_ros2/blob/main/ros2/test/pluginlib/BUILD.bazel#L67) in `pluginlib`. `ros2_py_binary` is almost the same as `ros2_py_test`, except not for tests, but "normal" executables.

thomasegriffith: Got it.  Thanks!

ros2/cc_defs.bzl
```diff
@@ -64,7 +64,39 @@ def ros2_c_binary(name, ros2_package_name = None, **kwargs):
...
+        data = [target_impl],
+        testonly = is_test,
+    )
+
+    sh_target = native.sh_test if is_test else native.sh_binary
+    sh_target(
+        name = name,
+        srcs = [launcher],
+        data = [target_impl],
+        tags = kwargs.pop("tags", []),
```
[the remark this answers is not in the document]
mvukov: AFAIK, if you would have tags in kwargs and also have manual tag for _ros2_cc_target, that would fail. The intention was to make sure that _ros2_cc_target only has manual tag. IMO, extraction of `tags` from kwargs should happen before the call to _ros2_cc_target.

ros2/cc_defs.bzl
```diff
@@ -64,7 +64,39 @@ def ros2_c_binary(name, ros2_package_name = None, **kwargs):
...
+        visibility = kwargs.pop("visibility", None),
```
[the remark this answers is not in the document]
mvukov: Fair enough (at least for now).

ros2/cc_defs.bzl
```diff
@@ -73,9 +105,10 @@ def ros2_cpp_binary(name, ros2_package_name = None, **kwargs):
         name: A unique target name.
         ros2_package_name: If given, defines a ROS package name for the target.
             Otherwise, the `name` is used as the package name.
-        **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-binaries
+        set_up_ament: If true, sets up ament file tree for the test target.
```
mvukov, line 108: `sets up ament file tree for the binary target.`

ros2/py_defs.bzl
```diff
@@ -3,48 +3,68 @@
...
     if set_up_ament == False:
-        py_test(name = name, **kwargs)
+        target(name = name, srcs = srcs, main = main, **kwargs)
         return
 
-    test_target = name + "_impl"
-    tags = kwargs.pop("tags", None)
-    py_test(name = test_target, tags = ["manual"], **kwargs)
+    target_impl = name + "_impl"
+    target(name = target_impl, srcs = srcs, main = main, tags = ["manual"], **kwargs)
```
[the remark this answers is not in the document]
mvukov: :+1: Indeed, this approach is just fine.

ros2/py_defs.bzl
```diff
@@ -3,48 +3,68 @@
...
 
-    native.sh_test(
+    sh_target = native.sh_test if is_test else native.sh_binary
+    sh_target(
         name = name,
         srcs = [launcher],
-        data = [test_target_symlink],
-        tags = tags,
+        data = [target_impl_symlink],
+        tags = kwargs.pop("tags", []),
```
mvukov, line 44: Same as for C++ macros: tags should be extracted from kwargs before the call to `target(.)`.

ros2/py_defs.bzl
```diff
@@ -3,48 +3,68 @@
...
     )
+
+def ros2_py_binary(name, srcs, main, set_up_ament = False, **kwargs):
+    """ Defines a ROS 2 Python binary.
+
+    Args:
+        name: A unique target name.
+        srcs: List of source files.
+        main: Source file to use as entrypoint.
+        set_up_ament: If true, sets up ament file tree for the test target.
```
mvukov, line 55: `sets up ament file tree for the binary target.`

ros2/test/pluginlib/BUILD.bazel
```diff
@@ -77,3 +77,50 @@ ros2_py_test(
     main = "py_loader_tests.py",
     set_up_ament = True,
 )
+
+ros2_cpp_binary(
```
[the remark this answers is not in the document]
mvukov: :+1:

ros2/py_defs.bzl
```diff
@@ -3,48 +3,68 @@
...
+        **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-tests
```
mvukov, line 56: https://bazel.build/reference/be/common-definitions#common-attributes-binaries

ros2/cc_defs.bzl
```diff
@@ -73,9 +105,10 @@ def ros2_cpp_binary(name, ros2_package_name = None, **kwargs):
...
+        **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-tests
```
mvukov, line 109: https://bazel.build/reference/be/common-definitions#common-attributes-binaries

mvukov (commented):
Overall approach is just fine, great work! I left some smaller comments.

ros2/cc_defs.bzl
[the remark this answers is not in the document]
ahans: Ah, yes, you're right! I reverted that change now.

ahans:
@mvukov I think I addressed all your comments. Please have another look!

ros2/py_defs.bzl
ahans: I accidentally changed the test rule to say binary! :facepalm: Good thing you had another close look. Thanks!

ros2/cc_defs.bzl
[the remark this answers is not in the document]
ahans: I agree. It's done locally, but since GH is experiencing issues currently, I can't push right now. Will ping you once you can have a (hopefully) final look.

ahans:
@mvukov  Now pushing worked and the changes are visible.

PR #93, opened 2023-03-27
Fix plugin registration with clang build

### LINKED ISSUE #78: Plugin registration doesn't work with clang compiler
If you have clang set up as default compiler, or explicitly force it in `.bazelrc` like
```
build --repo_env=CC=clang
build --repo_env=CXX=clang++
```

Then `bazel test //...` in this repo will fail.

Here is one test output that shows the issue:
```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //ros2/test/image_common:image_transport_plugins_tests
-----------------------------------------------------------------------------
Running main() from gmock_main.cc
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from TestImageTransportPlugins
[ RUN      ] TestImageTransportPlugins.WhenRawPluginsAvailable_EnsurePublisherPluginCanBeLoaded
unknown file: Failure
C++ exception with description "MultiLibraryClassLoader: Could not create object of class type image_transport::RawPublisher as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()" thrown in the test body.
[  FAILED  ] TestImageTransportPlugins.WhenRawPluginsAvailable_EnsurePublisherPluginCanBeLoaded (16 ms)
[ RUN      ] TestImageTransportPlugins.WhenRawPluginsAvailable_EnsureSubscriberPluginCanBeLoaded
unknown file: Failure
C++ exception with description "MultiLibraryClassLoader: Could not create object of class type image_transport::RawSubscriber as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()" thrown in the test body.
[  FAILED  ] TestImageTransportPlugins.WhenRawPluginsAvailable_EnsureSubscriberPluginCanBeLoaded (16 ms)
[----------] 2 tests from TestImageTransportPlugins (32 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test suite ran. (32 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] TestImageTransportPlugins.WhenRawPluginsAvailable_EnsurePublisherPluginCanBeLoaded
[  FAILED  ] TestImageTransportPlugins.WhenRawPluginsAvailable_EnsureSubscriberPluginCanBeLoaded

 2 FAILED TESTS
```

ahans:
`class_loader` has a couple of functions that return static variables to
manage global state. Most importantly, they are used by plugins (shared
objects loaded via `dlopen()`) to register themselves upon load. By
default, code living in a shared object cannot access symbols from the
main executable that loaded it. In rules_ros2, both main executables as
well as plugins linked `class_loader` statically. This lead to both
having a separate instance of `class_loader` that couldn't see each
other. So when loading a plugin, it would register in its own copy of
`class_loader`. To the main executable it looked like the plugin wasn't
exporting any classes. To fix this, we need to make sure that
`class_loader` is also loaded as a shared object. Then the same instance
is used from a main executable as well as other shared objects (i.e.,
plugins). We achieve this by using some Bazel trick where we build
`class_loader` only as a shared object (via `cc_binary`) and make that
available to the rest of the build via `cc_library` importing only that
shared object as `srcs`.

It is unclear how this was working with gcc before. It looks like the
class_loader patch that pulled the local static variables into some
single instance of a singleton class allowed using some gcc-specific
quirk (or bug?). Always dynamically linking `class_loader` makes the
clang build work and allows us to drop the patch.

Fixes #78.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,12 +27,20 @@ jobs:
     name: Run tests
     runs-on: ubuntu-20.04
 
-    # Run bazel test in each workspace
+    # Run bazel test with gcc and clang in each workspace
     strategy:
       matrix:
         folder:
           - "."
           - "examples"
+        toolchain:
+          - "gcc"
+          - "clang"
+        include:
+          - toolchain: "gcc"
+            config_option: ""
+          - toolchain: "clang"
+            config_option: "--config=clang"
 
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
@@ -68,12 +76,12 @@ jobs:
             ~/.cache/bazel-repo
           key: v3-bazel-repo-cache-${{ hashFiles('.bazelversion', 'examples/WORKSPACE', 'repositories/**', 'requirements_lock.txt', 'WORKSPACE') }}
           restore-keys: v3-bazel-repo-cache-
-      - name: bazel test //...
+      - name: bazel test ${{ matrix.config_option }} //...
         env:
           # Bazelisk will download bazel to here, ensure it is cached between runs.
           XDG_CACHE_HOME: ~/.cache/bazel-repo
         working-directory: ${{ matrix.folder }}
         run: |
           bazel --bazelrc=$GITHUB_WORKSPACE/.github/workflows/ci.bazelrc --bazelrc=.bazelrc \
-              test //... \
+              test ${{ matrix.config_option }} //... \
               --remote_header=x-buildbuddy-api-key=${{ secrets.BUILDBUDDY_ORG_API_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -68,12 +76,12 @@ jobs:
             ~/.cache/bazel-repo
           key: v3-bazel-repo-cache-${{ hashFiles('.bazelversion', 'examples/WORKSPACE', 'repositories/**', 'requirements_lock.txt', 'WORKSPACE') }}
```
ahans, line 77: Maybe we would want a separate cache per toolchain? Wasn't sure about that. In that case we'd have to include `${{ matrix.toolchain }}` in the `key` line.

mvukov: I don't think this is necessary, as those files are relevant for the repository cache, only.

mvukov (approved):
:100:

PR #100, opened 2023-04-21
Default ROS_HOME and ROS_LOG_DIR env vars in ros2_cpp_test

lalten:
Default `ROS_HOME` and `ROS_LOG_DIR` env vars in `ros2_cpp_test` to `TEST_UNDECLARED_OUTPUTS_DIR`.
This is of course inspired by https://github.com/mvukov/rules_ros2/blob/c3451871e6d74dda2fd247076b8e18c0e6c010c4/ros2/pytest_wrapper.py.tpl#L38-L39

`TEST_UNDECLARED_OUTPUTS_DIR` is optional(https://bazel.build/reference/test-encyclopedia#initial-conditions), only `TEST_TMPDIR` is required to be present. However I would argue that currently users have to set this manually already via the `env` attribute, so there just would not be any change for users where the var is not available.

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,2 +1,1 @@
 examples
-bazel-rules_ros2
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,1 +1,2 @@
 examples
+bazel-rules_ros2
--- a/ros2/cc_defs.bzl
+++ b/ros2/cc_defs.bzl
@@ -114,7 +114,7 @@ def ros2_cpp_binary(name, ros2_package_name = None, set_up_ament = False, **kwar
     """
     _ros2_cpp_exec(cc_binary, name, ros2_package_name, set_up_ament, **kwargs)
 
-def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = True, **kwargs):
+def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, **kwargs):
     """ Defines a ROS 2 C++ test.
 
     Adds common ROS 2 C++ definitions on top of a cc_test.
--- a/ros2/cc_defs.bzl
+++ b/ros2/cc_defs.bzl
@@ -124,4 +124,7 @@ def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, **kwargs
         set_up_ament: If true, sets up ament file tree for the test target.
         **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-tests
     """
-    _ros2_cpp_exec(cc_test, name, ros2_package_name, set_up_ament, **kwargs)
+    env = kwargs.pop("env", {})
+    env["ROS_HOME"] = env.get("ROS_HOME", "$TEST_UNDECLARED_OUTPUTS_DIR")
+    env["ROS_LOG_DIR"] = env.get("ROS_LOG_DIR", "$TEST_UNDECLARED_OUTPUTS_DIR")
+    _ros2_cpp_exec(cc_test, name, ros2_package_name, set_up_ament, env = env, **kwargs)
--- a/ros2/cc_defs.bzl
+++ b/ros2/cc_defs.bzl
@@ -64,8 +64,8 @@ def ros2_c_binary(name, ros2_package_name = None, **kwargs):
     """
     _ros2_cc_target(cc_binary, "c", name, ros2_package_name, **kwargs)
 
-def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False, set_up_ros_home = False, **kwargs):
-    if not (set_up_ament or set_up_ros_home):
+def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False, **kwargs):
+    if set_up_ament == False:
         _ros2_cc_target(target, "cpp", name, ros2_package_name, **kwargs)
         return
 
@@ -83,9 +83,7 @@ def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False,
         deps = [target_impl],
         template = "@com_github_mvukov_rules_ros2//ros2:launch.sh.tpl",
         substitutions = {
-            "{{set_up_ros_home}}": "set_up_ros_home" if set_up_ros_home else "",
-            "{{set_up_ament}}": "set_up_ament" if set_up_ament else "",
-            "{{entry_point}}": "$(rootpath {})".format(target_impl),
+            "{entry_point}": "$(rootpath {})".format(target_impl),
         },
         tags = ["manual"],
         data = [target_impl],
@@ -114,11 +112,9 @@ def ros2_cpp_binary(name, ros2_package_name = None, set_up_ament = False, **kwar
         set_up_ament: If true, sets up ament file tree for the binary target.
         **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-binaries
     """
-    if "set_up_ros_home" in kwargs:
-        fail("set_up_ros_home only makes sense for test targets.")
     _ros2_cpp_exec(cc_binary, name, ros2_package_name, set_up_ament, **kwargs)
 
-def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, set_up_ros_home = True, **kwargs):
+def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = True, **kwargs):
     """ Defines a ROS 2 C++ test.
 
     Adds common ROS 2 C++ definitions on top of a cc_test.
@@ -128,7 +124,6 @@ def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, set_up_r
         ros2_package_name: If given, defines a ROS package name for the target.
             Otherwise, the `name` is used as the package name.
         set_up_ament: If true, sets up ament file tree for the test target.
-        set_up_ros_home: If true, sets up ROS_HOME for the test target.
         **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-tests
     """
-    _ros2_cpp_exec(cc_test, name, ros2_package_name, set_up_ament, set_up_ros_home, **kwargs)
+    _ros2_cpp_exec(cc_test, name, ros2_package_name, set_up_ament, **kwargs)
--- a/ros2/cc_defs.bzl
+++ b/ros2/cc_defs.bzl
@@ -64,14 +64,15 @@ def ros2_c_binary(name, ros2_package_name = None, **kwargs):
     """
     _ros2_cc_target(cc_binary, "c", name, ros2_package_name, **kwargs)
 
-def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False, **kwargs):
-    if set_up_ament == False:
+def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False, set_up_ros_home = False, **kwargs):
+    if not (set_up_ament or set_up_ros_home):
         _ros2_cc_target(target, "cpp", name, ros2_package_name, **kwargs)
         return
 
     target_impl = name + "_impl"
     tags = kwargs.pop("tags", [])
     visibility = kwargs.pop("visibility", None)
+    size = kwargs.pop("size", None)
     _ros2_cc_target(cc_binary, "cpp", target_impl, ros2_package_name, tags = ["manual"], **kwargs)
 
     is_test = target == cc_test
@@ -82,7 +83,9 @@ def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False,
         deps = [target_impl],
         template = "@com_github_mvukov_rules_ros2//ros2:launch.sh.tpl",
         substitutions = {
-            "{entry_point}": "$(rootpath {})".format(target_impl),
+            "{{set_up_ros_home}}": "set_up_ros_home" if set_up_ros_home else "",
+            "{{set_up_ament}}": "set_up_ament" if set_up_ament else "",
+            "{{entry_point}}": "$(rootpath {})".format(target_impl),
         },
         tags = ["manual"],
         data = [target_impl],
@@ -92,6 +95,7 @@ def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False,
     sh_target = native.sh_test if is_test else native.sh_binary
     sh_target(
         name = name,
+        size = size,
         srcs = [launcher],
         data = [target_impl],
         tags = tags,
@@ -110,9 +114,11 @@ def ros2_cpp_binary(name, ros2_package_name = None, set_up_ament = False, **kwar
         set_up_ament: If true, sets up ament file tree for the binary target.
         **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-binaries
     """
+    if "set_up_ros_home" in kwargs:
+        fail("set_up_ros_home only makes sense for test targets.")
     _ros2_cpp_exec(cc_binary, name, ros2_package_name, set_up_ament, **kwargs)
 
-def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, **kwargs):
+def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, set_up_ros_home = True, **kwargs):
     """ Defines a ROS 2 C++ test.
 
     Adds common ROS 2 C++ definitions on top of a cc_test.
@@ -122,9 +128,7 @@ def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, **kwargs
         ros2_package_name: If given, defines a ROS package name for the target.
             Otherwise, the `name` is used as the package name.
         set_up_ament: If true, sets up ament file tree for the test target.
+        set_up_ros_home: If true, sets up ROS_HOME for the test target.
         **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-tests
     """
-    env = kwargs.pop("env", {})
-    env["ROS_HOME"] = env.get("ROS_HOME", "$TEST_UNDECLARED_OUTPUTS_DIR")
-    env["ROS_LOG_DIR"] = env.get("ROS_LOG_DIR", "$TEST_UNDECLARED_OUTPUTS_DIR")
-    _ros2_cpp_exec(cc_test, name, ros2_package_name, set_up_ament, env = env, **kwargs)
+    _ros2_cpp_exec(cc_test, name, ros2_package_name, set_up_ament, set_up_ros_home, **kwargs)
--- a/ros2/launch.sh.tpl
+++ b/ros2/launch.sh.tpl
@@ -2,21 +2,17 @@
 
 set -o errexit -o nounset -o pipefail
 
-if [ -n "{{set_up_ros_home}}" ]; then
-  if [ -z "${ROS_HOME:-}" ] && [ -z "${ROS_LOG_DIR:-}" ]; then
-    ros_output_dir="${TEST_UNDECLARED_OUTPUTS_DIR:-${TEST_TMPDIR:-}}"
-    if [ -n "${ros_output_dir}" ]; then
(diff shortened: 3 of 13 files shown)


mvukov:
Your observation regarding TEST_UNDECLARED_OUTPUTS_DIR is spot-on. I would leave Bazel macro as-is, and change the templates (test.py.tpl and pytest_wrapper.py.tpl) directly instead. E.g.
```py
test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
if test_outputs_dir is None:
    test_outputs_dir = os.environ.get('TEST_TMPDIR')
if test_outputs_dir is None:
    test_outputs_dir = os.getcwd()

os.environ['ROS_HOME'] = test_outputs_dir
os.environ['ROS_LOG_DIR'] = test_outputs_dir
```
The second `if` may still be needed if folks run tests with `bazel run`.

The rclcpp test is fine for me.

lalten:
> leave Bazel macro as-is, and change the templates

Thanks for the feedback! This was a lot more difficult than I anticipated at first :D I wanted to make the setting up ROS_HOME in tests the default, but it wasn't super straightforward to just default set_up_ament to True to actually use the wrapper for all tests: There are some `$(rootpath //pkg)` that then would become `$(rootpaths //pkg)`, which didn't work out.

Anyways, let me know what you think about this new version :)

ros2/cc_defs.bzl
```diff
@@ -64,14 +64,15 @@ def ros2_c_binary(name, ros2_package_name = None, **kwargs):
...
-    if set_up_ament == False:
+def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False, set_up_ros_home = False, **kwargs):
+    if not (set_up_ament or set_up_ros_home):
         _ros2_cc_target(target, "cpp", name, ros2_package_name, **kwargs)
         return
 
     target_impl = name + "_impl"
     tags = kwargs.pop("tags", [])
     visibility = kwargs.pop("visibility", None)
+    size = kwargs.pop("size", None)
```
mvukov, line 75: OK, nice, I overlooked this.

ros2/cc_defs.bzl
```diff
@@ -92,6 +95,7 @@ def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False,
     sh_target = native.sh_test if is_test else native.sh_binary
     sh_target(
         name = name,
+        size = size,
```
mvukov, line 98: Not sure this will work with a sh_binary. Can you double-check? An alternative is to create a separate kwargs-like dict that would be empty for sh_binary and have size for sh_test. That dict you would add to the args instead of the size directly.

ros2/test.py.tpl
```diff
@@ -17,7 +17,7 @@ sys.argv = sys.argv[:1] + [
     LAUNCH_FILE,
 ] + sys.argv[1:]
 
-test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
+test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR') or os.environ.get('TEST_TMPDIR')
```
mvukov, line 20: https://google.github.io/styleguide/pyguide.html#2144-decision: get() can return None and one should check that as `if foo is None:`

.bazelignore
```diff
@@ -1 +1,2 @@
 examples
+bazel-rules_ros2
```
mvukov, line 2: I guess this is for debugging. Please rm.

mvukov (commented):
OK, I think you want to be able to override ROS_HOME and ROS_LOG_DIR, even outside the sandbox.

I think the following would do the job in templates (test.py.tpl and pytest_wrapper.py.tpl):
```py
test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
if test_outputs_dir is None:
    test_outputs_dir = os.environ.get('TEST_TMPDIR')

if os.environ.get('ROS_HOME') is not None:
    os.environ['ROS_HOME'] = test_outputs_dir

if os.environ.get('ROS_LOG_DIR') is not None:
    os.environ['ROS_LOG_DIR'] = test_outputs_dir
```
then you can override the both with e.g. `--test_env=ROS_HOME=/my/custom/path`. This looks to me as a very general approach and avoids a lot of other changes in this PR.

ros2/cc_defs.bzl
lalten: Looks like sh_binary is fine with it, although tbh I can't immediately find an explaanation for it in https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/sh/BazelShBinaryRule.java

I added a print in just before the sh_target:
```starlark
    if not is_test:
        print("{} is a {} with size={}".format(name, sh_target, size))
```
and it will print
```
DEBUG: (...)/rules_ros2/ros2/cc_defs.bzl:97:14: plugin_tests_binary is a <built-in rule sh_binary> with size=None
```

.bazelignore
lalten: You're right -- This is needed to make the Starlark debugger in VS Code work, see https://github.com/bazelbuild/bazel/issues/10653#issuecomment-694230015
it's of course unrelated to the actual changes in the PR, so I removed it again

ros2/test.py.tpl
lalten: Thx, didn't know the Google Styleguide disallows `or` short-circuiting!

lalten:
> Please take a look at my last comment, where I suggested another setup of relevant ROS env vars.

This is in the templates now:
```py
    test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
    if test_outputs_dir is None:
        test_outputs_dir = os.environ.get('TEST_TMPDIR')
    if test_outputs_dir:
        if 'ROS_HOME' not in os.environ and 'ROS_LOG_DIR' not in os.environ:
            os.environ['ROS_HOME'] = test_outputs_dir
            os.environ['ROS_LOG_DIR'] = test_outputs_dir
```

The fallback from `TEST_UNDECLARED_OUTPUTS_DIR` to `TEST_TMPDIR` is as you suggested.

What's different is that I set both `ROS_HOME` and `ROS_LOG_DIR` iff neither is set. The reason is that if either one of them is set, rclcpp (rclpy) will work (ROS_LOG_DIR defaults to `$ROS_HOME/.log`). The issue I'm trying to fix will only appear if both env vars are unset. The reason I'm setting both, and to the same value, is that this is the current behavior in the pytest wrapper. If it's ok to break backwards compatibility (which we could certainly consider) it would be ok to set ROS_HOME only.

ros2/cc_defs.bzl
```diff
@@ -64,14 +64,15 @@ def ros2_c_binary(name, ros2_package_name = None, **kwargs):
     """
     _ros2_cc_target(cc_binary, "c", name, ros2_package_name, **kwargs)
 
-def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False, **kwargs):
-    if set_up_ament == False:
+def _ros2_cpp_exec(target, name, ros2_package_name = None, set_up_ament = False, set_up_ros_home = False, **kwargs):
```
mvukov, line 67: Why do we still need set_up_ros_home when the templates are updated?

ros2/test.py.tpl
```diff
@@ -18,9 +18,12 @@ sys.argv = sys.argv[:1] + [
 ] + sys.argv[1:]
 
 test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
+if test_outputs_dir is None:
+    test_outputs_dir = os.environ.get('TEST_TMPDIR')
 if test_outputs_dir:
```
mvukov, line 23: This is not needed as per https://bazel.build/reference/test-encyclopedia#initial-conditions TEST_TMPDIR is always available.

ros2/test.py.tpl
```diff
@@ -18,9 +18,12 @@ sys.argv = sys.argv[:1] + [
...
-    os.environ['ROS_HOME'] = test_outputs_dir
-    os.environ['ROS_LOG_DIR'] = test_outputs_dir
+    if 'ROS_HOME' not in os.environ and 'ROS_LOG_DIR' not in os.environ:
+        os.environ['ROS_HOME'] = test_outputs_dir
+        os.environ['ROS_LOG_DIR'] = test_outputs_dir
```
mvukov, line 26: Why not treat them separately as I suggested?

ros2/pytest_wrapper.py.tpl
```diff
@@ -34,9 +34,12 @@ def finalize_coverage_session(coverage_session: coverage.Coverage) -> None:
...
 def main() -> None:
     test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
+    if test_outputs_dir is None:
+        test_outputs_dir = os.environ.get('TEST_TMPDIR')
     if test_outputs_dir:
-        os.environ['ROS_HOME'] = test_outputs_dir
-        os.environ['ROS_LOG_DIR'] = test_outputs_dir
+        if 'ROS_HOME' not in os.environ and 'ROS_LOG_DIR' not in os.environ:
+            os.environ['ROS_HOME'] = test_outputs_dir
+            os.environ['ROS_LOG_DIR'] = test_outputs_dir
```
mvukov, line 42: Same comments as for the other template.

ros2/launch.sh.tpl
```diff
@@ -2,10 +2,21 @@
...
-if [ -z "${ament_prefix_path}" ]; then
+if [ -n "{{set_up_ros_home}}" ]; then
+  if [ -z "${ROS_HOME:-}" ] && [ -z "${ROS_LOG_DIR:-}" ]; then
+    ros_output_dir="${TEST_UNDECLARED_OUTPUTS_DIR:-${TEST_TMPDIR:-}}"
+    if [ -n "${ros_output_dir}" ]; then
+      export ROS_HOME="${ros_output_dir}"
+      export ROS_LOG_DIR="${ros_output_dir}"
+    fi
+  fi
+fi
```
mvukov, line 13: OK, now I think I understand what you want to do. According to the test encyclopedia BAZEL_TEST is always defined when users do `bazel test`. So, I would do the following

```sh
if BAZEL_TEST defined:
  <reimplement python logic for setting up ROS home and log dirs>
  # NOTE: TEST_TMPDIR is always defined.
  export ROS_HOME=...
  export ROS_LOG_DIR=...
```

ros2/launch.sh.tpl
```diff
@@ -2,10 +2,21 @@
...
+
+if [ -n "{{set_up_ament}}" ]; then
```
mvukov, line 15: I don't think you need set_up_ament and the code below can be reverted IMO.

mvukov (commented):
IMO, `set_up_ros_home` arg in set_up_ros_home can be avoided. Please take a look at my comments.

ros2/test.py.tpl
lalten: good point!

ros2/cc_defs.bzl
lalten: The idea was to keep setting up ament and setting up ROS_HOME orthogonal. We could change the flag to do both of course

ros2/test.py.tpl
lalten: tbh I think the best way would be to set only ROS_HOME when neither ROS_HOME nor ROS_LOG_DIR is set. But as I said, this isn't the current behavior of the pytest wrapper template, and it seems like a good idea to be consistent with that

lalten:
I've simplified the wrapper setup to remove the setup_ros_home arg, and changed the tests to make sure each of the three test drivers is verified:

- `ros2_cpp_test`: `launch.sh.tpl`
- `ros2_py_test`: `launch.sh.tpl`
- `ros2_test` with `use_pytest = True`: `launch.py.tpl`
- `ros2_test` with `use_pytest = False`: `pytest_wrapper.py.tpl`

lalten:
Btw, I think something is wrong with the CI setup, maybe you could check?
```
ERROR: The Build Event Protocol upload failed: Not retrying publishBuildEvents, no more attempts left: status='Status{code=UNAUTHENTICATED, description=failed to parse API key: missing API Key, cause=null}' UNAUTHENTICATED: UNAUTHENTICATED: failed to parse API key: missing API Key UNAUTHENTICATED: UNAUTHENTICATED: failed to parse API key: missing API Key
```

mvukov:
https://github.com/mvukov/rules_ros2/pull/102/files might make problems... I need to investigate this further. The API key for buildbuddy remote cache is stored as a GH secret and it might be that with the new config it's not not available any more -- hence, the error.

You can try to revert that in your branch and see if the CI works then.

mvukov:
Just took a look at https://github.com/aspect-build/rules_js/blob/main/.github/workflows/ci.yaml. I need to fix CI config to work on pull requests from forks w/o github secrets.

ros2/launch.sh.tpl
```diff
@@ -2,6 +2,13 @@
 
 set -o errexit -o nounset -o pipefail
 
+if [ -z "${ROS_HOME:-}" ] && [ -z "${ROS_LOG_DIR:-}" ]; then
```
mvukov, line 5: For relation(s) between ROS_HOME and ROS_LOG_DIR folks should read ROS docs, not docs/impl of this repo. Here we should tend to be as general as possible and avoid introducing hidden functionality. Therefore, I'd ask again for separate setup of those two env vars like I explained earlier.

Next to this, this piece of code is only relevant for tests. Thus, this should look like (for better explicitness IMO):
```sh
if [[ ! -z "${BAZEL_TEST}" ]]; then
  bazel_test_output_dir="${TEST_UNDECLARED_OUTPUTS_DIR:-${TEST_TMPDIR}}"
  if [[ -z "${ROS_HOME:-}" ]]; then
    export ROS_HOME="${bazel_test_output_dir}" 
  fi
  if [[ -z "${ROS_LOG_DIR:-}" ]]; then
    export ROS_LOG_DIR="${bazel_test_output_dir}" 
  fi
fi
```

ros2/py_defs.bzl
```diff
@@ -59,7 +61,7 @@ def ros2_py_binary(name, srcs, main, set_up_ament = False, **kwargs):
     """
     _ros2_py_exec(py_binary, name, srcs, main, set_up_ament, **kwargs)
 
-def ros2_py_test(name, srcs, main, set_up_ament = False, **kwargs):
+def ros2_py_test(name, srcs, main, set_up_ament = True, **kwargs):
```
mvukov, line 64: Same here: this is for a follow-up PR.

ros2/pytest_wrapper.py.tpl
```diff
@@ -33,10 +33,12 @@ def finalize_coverage_session(coverage_session: coverage.Coverage) -> None:
...
-    test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
-    if test_outputs_dir:
-        os.environ['ROS_HOME'] = test_outputs_dir
-        os.environ['ROS_LOG_DIR'] = test_outputs_dir
+    if 'ROS_HOME' not in os.environ and 'ROS_LOG_DIR' not in os.environ:
+        ros_output_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
+        if ros_output_dir is None:
+            ros_output_dir = os.environ.get('TEST_TMPDIR')
+        if ros_output_dir is not None:
+            os.environ['ROS_HOME'] = ros_output_dir
```
mvukov, line 41: Please use a snippet I posted earlier. Same reasoning as for launch.sh.tpl.

ros2/test.py.tpl
```diff
@@ -17,10 +17,12 @@ sys.argv = sys.argv[:1] + [
     LAUNCH_FILE,
 ] + sys.argv[1:]
 
-test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
-if test_outputs_dir:
-    os.environ['ROS_HOME'] = test_outputs_dir
-    os.environ['ROS_LOG_DIR'] = test_outputs_dir
+if 'ROS_HOME' not in os.environ and 'ROS_LOG_DIR' not in os.environ:
```
mvukov, line 20: Please use a snippet I posted earlier.

ros2/test/BUILD.bazel
```diff
@@ -1,11 +1,9 @@
...
 """
 
-load(
-    "@com_github_mvukov_rules_ros2//ros2:interfaces.bzl",
-    "py_ros2_interface_library",
-    "ros2_interface_library",
-)
+load("@com_github_mvukov_rules_ros2//ros2:cc_defs.bzl", "ros2_cpp_test")
+load("@com_github_mvukov_rules_ros2//ros2:interfaces.bzl", "py_ros2_interface_library", "ros2_interface_library")
+load("@com_github_mvukov_rules_ros2//ros2:py_defs.bzl", "ros2_py_test")
```
mvukov, line 6: Not needed.

ros2/test/BUILD.bazel
```diff
@@ -1,11 +1,9 @@
 """ Defines tests.
 """
 
-load(
-    "@com_github_mvukov_rules_ros2//ros2:interfaces.bzl",
-    "py_ros2_interface_library",
-    "ros2_interface_library",
-)
+load("@com_github_mvukov_rules_ros2//ros2:cc_defs.bzl", "ros2_cpp_test")
+load("@com_github_mvukov_rules_ros2//ros2:interfaces.bzl", "py_ros2_interface_library", "ros2_interface_library")
```
mvukov, line 5: Please revert formatting as in original.

ros2/pytest_wrapper.py.tpl
```diff
@@ -33,10 +33,12 @@ def finalize_coverage_session(coverage_session: coverage.Coverage) -> None:
...
 def main() -> None:
-    test_outputs_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
-    if test_outputs_dir:
-        os.environ['ROS_HOME'] = test_outputs_dir
-        os.environ['ROS_LOG_DIR'] = test_outputs_dir
+    if 'ROS_HOME' not in os.environ and 'ROS_LOG_DIR' not in os.environ:
+        ros_output_dir = os.environ.get('TEST_UNDECLARED_OUTPUTS_DIR')
+        if ros_output_dir is None:
+            ros_output_dir = os.environ.get('TEST_TMPDIR')
+        if ros_output_dir is not None:
```
mvukov, line 40: TEST_TMPDIR is always defined, we don't need this `if-`clause here. The previous line should read: `ros_output_dir = os.environ['TEST_TMPDIR']`.

BTW, I would rename ros_output_dir to bazel_test_output_dir for more explicitness.

ros2/test/test_rclcpp.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+int main(int argc, char** argv) {
+  // If neither $ROS_HOME nor $ROS_LOG_DIR are set to a valid directory,
+  // rclcpp:init() will fail with these errors:
+  // clang-format off
+  // 'rcutils_expand_user failed, at external/ros2_rcl_logging/rcl_logging_interface/src/logging_dir.c:81'             // NOLINT
+  // 'Failed to get logging directory, at external/ros2_rcl_logging/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp:83'  // NOLINT
+  // clang-format on
+  rclcpp::init(argc, argv);
+  return 0;
```
mvukov, line 25: Maybe `return rclcpp::shutdown()`?

mvukov (changes requested):
To summarize, this PR addresses 3 issues:
1. Test size routing in custom test macros was missing, this is fixed now.
2. Without proper handling of ROS_HOME and ROS_LOGGING_DIR, some tests could escape the sandbox. This PR will fix that.
3. Besides this, it looks like there are cases when users want to set ROS_HOME and ROS_LOGGING_DIR explicitly in tests. This PR will also allow that.

mvukov (approved):

mvukov:
The PR is in good shape now IMO. Thanks for your efforts!

PR #105, opened 2023-05-08
Default `set_up_ament` to `True` for `ros2_cpp_test`, `ros2_py_test`

lalten:
I believe this is a reasonable default because it shouldn't affect the observable behavior of existing tests, but will save new users from some debugging when they try to figure out what is causing the `rcutils_expand_user failed`, `Failed to get logging directory` errors in their tests that spin up the RCL.

### DIFF
--- a/ros2/cc_defs.bzl
+++ b/ros2/cc_defs.bzl
@@ -109,12 +109,13 @@ def ros2_cpp_binary(name, ros2_package_name = None, set_up_ament = False, **kwar
         name: A unique target name.
         ros2_package_name: If given, defines a ROS package name for the target.
             Otherwise, the `name` is used as the package name.
-        set_up_ament: If true, sets up ament file tree for the binary target.
+        set_up_ament: If true, wrap the binary target to set up ament file tree and
+            missing ROS_HOME/ROS_LOG_DIR env vars.
         **kwargs: https://bazel.build/reference/be/common-definitions#common-attributes-binaries
     """
     _ros2_cpp_exec(cc_binary, name, ros2_package_name, set_up_ament, **kwargs)
 
-def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, **kwargs):
+def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = True, **kwargs):
     """ Defines a ROS 2 C++ test.
 
     Adds common ROS 2 C++ definitions on top of a cc_test.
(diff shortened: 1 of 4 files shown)


ros2/cc_defs.bzl
```diff
@@ -109,12 +109,13 @@ def ros2_cpp_binary(name, ros2_package_name = None, set_up_ament = False, **kwar
         name: A unique target name.
         ros2_package_name: If given, defines a ROS package name for the target.
             Otherwise, the `name` is used as the package name.
-        set_up_ament: If true, sets up ament file tree for the binary target.
+        set_up_ament: If true, wrap the binary target to set up ament file tree and
+            missing ROS_HOME/ROS_LOG_DIR env vars.
```
mvukov, line 113: Setting up ROS_HOME+ROS_LOG_DIR is only needed for tests, IMO. This comment can be reverted. In addition, the logic in launch.sh.tpl only sets up those env vars for tests.

ros2/py_defs.bzl
```diff
@@ -56,19 +56,21 @@ def ros2_py_binary(name, srcs, main, set_up_ament = False, **kwargs):
         name: A unique target name.
         srcs: List of source files.
         main: Source file to use as entrypoint.
-        set_up_ament: If true, sets up ament file tree for the binary target.
+        set_up_ament: If true, wrap the binary target to set up ament file tree and
```
mvukov, line 59: Same here.

ros2/cc_defs.bzl
```diff
@@ -123,7 +124,8 @@ def ros2_cpp_test(name, ros2_package_name = None, set_up_ament = False, **kwargs
         name: A unique target name.
         ros2_package_name: If given, defines a ROS package name for the target.
             Otherwise, the `name` is used as the package name.
-        set_up_ament: If true, sets up ament file tree for the test target.
+        set_up_ament: If true, wrap the test target to set up ament file tree and
```
mvukov, line 127: Please rephrase here and for the corr. Python macro as e.g.

```
If true, a custom launcher is generated that sets up the following for the test target:
* the ament file tree
* ROS_HOME and ROS_LOG_DIR are set to TEST_UNDECLARED_OUTPUTS_DIR, if available, otherwise to TEST_TMPDIR. See [here](https://bazel.build/reference/test-encyclopedia#initial-conditions) for more info.
```

mvukov:
@ahans PTAL.

ros2/cc_defs.bzl
lalten: true

ahans (approved):
LGTM

PR #107, opened 2023-05-09
Upgrade to release-humble-20230502

lalten:
I was mostly curious about how the `bazel run //repositories/private:resolver` worked (nice one!), and what better way to try it than to actually upgrade the Humble patch release.
I checked and none of the patches are obsolete, looks like the ones that are merged upstream were not backported from rolling to humble.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Available features:
 
 Please take a look at the [examples](examples) folder to get started.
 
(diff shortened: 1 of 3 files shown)


mvukov (approved):
Very nice! It was on my TODO list, but you were faster :)
Yeah, it's a patch release, as can be seen from the updates to the _impl.bzl file. I wouldn't expect significant changes TBH -- it's good that there's almost no work on our end :).

PR #108, opened 2023-05-10
Forward specific launcher_target_attrs to the wrapper

lalten:
Turns out I needed `args` as well on one test... I figured at this point it may be nicer to generalize :)

### DIFF
--- a/ros2/ament.bzl
+++ b/ros2/ament.bzl
@@ -319,6 +319,12 @@ def py_launcher(name, deps, idl_deps = None, **kwargs):
     )
     return name + ".py"
 
+def split_kwargs(**kwargs):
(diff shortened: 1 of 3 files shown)


mvukov (commented):
Very nice. How about extracting a helper function in ament.bzl (since the launcher logic is mostly therein)? For instance:
```python
def extract_launcher_kwargs(**kwargs):
    launcher_attrs = ["args", "size", "tags", "timeout", "visibility"]
    return {attr: kwargs.pop(attr) for attr in launcher_attrs if attr in kwargs}
```
Then the code dup will be eliminated.

mvukov (approved):

PR #111, opened 2023-05-15
Add a method to setup AMENT_PREFIX_PATH in C++

woutervdstoel:
I have a use case where it is not desirable to have a sh wrapper around my binary to set up the `AMENT_PREFIX_PATH` environment variable. This PR adds a way to set the variable from C++ code instead.

### DIFF
--- a/ros2/BUILD.bazel
+++ b/ros2/BUILD.bazel
@@ -5,6 +5,8 @@ load("@rules_python//python:defs.bzl", "py_binary", "py_library")
 
 exports_files([
     "ament.bzl",
+    "ament_setup.cc.tpl",
+    "ament_setup.h.tpl",
     "bag.bzl",
     "bag.py.tpl",
     "cc_defs.bzl",
@@ -21,8 +23,6 @@ exports_files([
     "test.bzl",
     "test.py.tpl",
     "topic.bzl",
-    "ament_setup.cc.tpl",
-    "ament_setup.h.tpl",
 ])
 
 py_library(
--- a/ros2/BUILD.bazel
+++ b/ros2/BUILD.bazel
@@ -21,6 +21,8 @@ exports_files([
     "test.bzl",
     "test.py.tpl",
     "topic.bzl",
+    "ament_setup.cpp.tpl",
+    "ament_setup.hpp.tpl",
 ])
 
 py_library(
--- a/ros2/BUILD.bazel
+++ b/ros2/BUILD.bazel
@@ -21,8 +21,8 @@ exports_files([
     "test.bzl",
     "test.py.tpl",
     "topic.bzl",
-    "ament_setup.cpp.tpl",
-    "ament_setup.hpp.tpl",
+    "ament_setup.cc.tpl",
+    "ament_setup.h.tpl",
 ])
 
 py_library(
--- a/ros2/ament.bzl
+++ b/ros2/ament.bzl
@@ -393,14 +393,15 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
     )
 
 def _cpp_ament_setup_impl(ctx):
-    src_output = ctx.actions.declare_file(ctx.attr.basename + "/ament_setup.cc")
-    hdr_output = ctx.actions.declare_file(ctx.attr.basename + "/ament_setup.h")
+    basename = ctx.attr.basename or ctx.label.name
+    src_output = ctx.actions.declare_file(basename + "/ament_setup.cc")
+    hdr_output = ctx.actions.declare_file(basename + "/ament_setup.h")
     ament_prefix_path = ctx.attr.ament_setup[Ros2AmentSetupInfo].ament_prefix_path
 
     substitutions = {
         "{{ament_prefix_path}}": ament_prefix_path,
         "{{header}}": hdr_output.short_path,
-        "{{namespace}}": ctx.attr.basename,
+        "{{namespace}}": basename,
     }
 
     expand_template_impl(
@@ -432,7 +433,7 @@ cpp_ament_setup = rule(
             providers = [Ros2AmentSetupInfo],
         ),
         "basename": attr.string(
-            mandatory = True,
+            default = "",
         ),
         "data": attr.label_list(allow_files = True),
         "_hdr_template": attr.label(
@@ -451,9 +452,14 @@ def cpp_ament_setup_library(name, deps, idl_deps = None, **kwargs):
     """ Generates a C++ library that contains the ament setup code for a ROS2 package.
 
     This can be useful in contexts where using a wrapper script (as is the case with the regular ros2_xxx rules)
-    is not desirable. The generated function can be called with an optional allow_append argument, which causes the
+    is not desirable. The generated library can be included as `#include <package_name>/<name>/ament_setup.h` and
+    contains a single function called `SetUpAmentPrefixPath` in the namespace `<name>`. That function needs to be
+    called before any ament resource is used.
+
+    The generated function can be called with an optional boolean allow_append argument, which causes the
     function to append to the AMENT_PREFIX_PATH instead of overwriting it. Beware that when using this mode when
-    different versions of the same plugin exist, the behavior might be unexpected.
+    a different version of the same plugin may exist in the existing AMENT_PREFIX_PATH, the behavior might be
+    unexpected.
 
     Args:
         name: The name of the target.
--- a/ros2/ament.bzl
+++ b/ros2/ament.bzl
@@ -488,7 +488,6 @@ def cpp_ament_setup_library(name, deps, idl_deps = None, **kwargs):
     cc_library(
         name = name,
         srcs = [":{}_srcs".format(name)],
-        deps = [],
         data = [ament_setup],
         testonly = testonly,
         **kwargs
--- a/ros2/ament.bzl
+++ b/ros2/ament.bzl
@@ -392,29 +392,27 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
         **kwargs
     )
 
-def _ros2_ament_cpp_library_impl(ctx):
-    src_output = ctx.actions.declare_file(ctx.attr.basename + ".cpp")
-    hdr_output = ctx.actions.declare_file(ctx.attr.basename + ".hpp")
-    ament_prefix_path = ctx.attr.ament_setup[Ros2AmentSetupInfo].ament_prefix_path or ""
+def _cpp_ament_setup_impl(ctx):
+    src_output = ctx.actions.declare_file(ctx.attr.basename + "/ament_setup.cc")
+    hdr_output = ctx.actions.declare_file(ctx.attr.basename + "/ament_setup.h")
+    ament_prefix_path = ctx.attr.ament_setup[Ros2AmentSetupInfo].ament_prefix_path
 
-    substitutions = dicts.add(
-        ctx.attr.substitutions,
-        {
-            "{{ament_prefix_path}}": ament_prefix_path,
-            "{{header}}": hdr_output.short_path,
-        },
-    )
+    substitutions = {
+        "{{ament_prefix_path}}": ament_prefix_path,
+        "{{header}}": hdr_output.short_path,
+        "{{namespace}}": ctx.attr.basename,
+    }
 
     expand_template_impl(
         ctx,
-        template = ctx.file.src_template,
+        template = ctx.file._src_template,
         output = src_output,
         substitutions = substitutions,
         is_executable = False,
     )
     expand_template_impl(
         ctx,
-        template = ctx.file.hdr_template,
+        template = ctx.file._hdr_template,
         output = hdr_output,
         substitutions = substitutions,
         is_executable = False,
@@ -427,28 +425,29 @@ def _ros2_ament_cpp_library_impl(ctx):
         ),
     ]
 
-ros2_ament_cpp_library_src = rule(
+cpp_ament_setup = rule(
     attrs = {
         "ament_setup": attr.label(
             mandatory = True,
             providers = [Ros2AmentSetupInfo],
         ),
-        "basename": attr.string(mandatory = True),
+        "basename": attr.string(
+            mandatory = True,
+        ),
         "data": attr.label_list(allow_files = True),
-        "hdr_template": attr.label(
-            mandatory = True,
+        "_hdr_template": attr.label(
             allow_single_file = True,
+            default = "@com_github_mvukov_rules_ros2//ros2:ament_setup.h.tpl",
         ),
-        "src_template": attr.label(
-            mandatory = True,
+        "_src_template": attr.label(
             allow_single_file = True,
+            default = "@com_github_mvukov_rules_ros2//ros2:ament_setup.cc.tpl",
         ),
-        "substitutions": attr.string_dict(mandatory = True),
     },
-    implementation = _ros2_ament_cpp_library_impl,
+    implementation = _cpp_ament_setup_impl,
 )
 
-def ros2_ament_setup_library(name, deps, idl_deps = None, **kwargs):
+def cpp_ament_setup_library(name, deps, idl_deps = None, **kwargs):
     ament_setup = name + "_ament_setup"
     testonly = kwargs.get("testonly", False)
     ros2_ament_setup(
@@ -460,13 +459,11 @@ def ros2_ament_setup_library(name, deps, idl_deps = None, **kwargs):
         testonly = testonly,
     )
     cpp_namespace = native.package_name().replace("/", "_") + "_" + name
-    ros2_ament_cpp_library_src(
+    cpp_ament_setup(
         name = name + "_srcs",
         basename = name,
         ament_setup = ament_setup,
-        src_template = "@com_github_mvukov_rules_ros2//ros2:ament_setup.cpp.tpl",
-        hdr_template = "@com_github_mvukov_rules_ros2//ros2:ament_setup.hpp.tpl",
-        substitutions = {"{{namespace}}": cpp_namespace},
+        tags = ["manual"],
     )
 
     cc_library(
--- a/ros2/ament.bzl
+++ b/ros2/ament.bzl
@@ -448,6 +448,19 @@ cpp_ament_setup = rule(
 )
 
 def cpp_ament_setup_library(name, deps, idl_deps = None, **kwargs):
+    """ Generates a C++ library that contains the ament setup code for a ROS2 package.
+
+    This can be useful in contexts where using a wrapper script (as is the case with the regular ros2_xxx rules)
+    is not desirable. The generated function can be called with an optional allow_append argument, which causes the
+    function to append to the AMENT_PREFIX_PATH instead of overwriting it. Beware that when using this mode when
+    different versions of the same plugin exist, the behavior might be unexpected.
+
+    Args:
+        name: The name of the target.
+        deps: The dependencies, typically targets of type ros2_plugin.
+        idl_deps: any IDL deps.
+        kwargs: Forwarded to the cc_library rule.
+    """
     ament_setup = name + "_ament_setup"
     testonly = kwargs.get("testonly", False)
     ros2_ament_setup(
@@ -458,7 +471,7 @@ def cpp_ament_setup_library(name, deps, idl_deps = None, **kwargs):
         tags = ["manual"],
         testonly = testonly,
     )
-    cpp_namespace = native.package_name().replace("/", "_") + "_" + name
+
     cpp_ament_setup(
         name = name + "_srcs",
         basename = name,
--- a/ros2/ament.bzl
+++ b/ros2/ament.bzl
@@ -23,6 +23,7 @@ load(
     "@com_github_mvukov_rules_ros2//third_party:expand_template.bzl",
     "expand_template_impl",
 )
+load("@rules_cc//cc:defs.bzl", "cc_library")
 
 _AMENT_SETUP_MODULE = "ament_setup"
 
@@ -291,12 +292,12 @@ py_launcher_rule = rule(
             mandatory = True,
             providers = [Ros2AmentSetupInfo],
         ),
+        "data": attr.label_list(allow_files = True),
+        "substitutions": attr.string_dict(mandatory = True),
         "template": attr.label(
             mandatory = True,
             allow_single_file = True,
         ),
-        "substitutions": attr.string_dict(mandatory = True),
-        "data": attr.label_list(allow_files = True),
     },
     implementation = _py_launcher_rule_impl,
 )
@@ -334,8 +335,8 @@ def _sh_launcher_rule_impl(ctx):
     substitutions = dicts.add(
         ctx.attr.substitutions,
         {
-            "{{bash_bin}}": ctx.toolchains[SH_TOOLCHAIN].path,
             "{{ament_prefix_path}}": ament_prefix_path,
+            "{{bash_bin}}": ctx.toolchains[SH_TOOLCHAIN].path,
         },
     )
 
@@ -363,12 +364,12 @@ sh_launcher_rule = rule(
             mandatory = True,
             providers = [Ros2AmentSetupInfo],
         ),
+        "data": attr.label_list(allow_files = True),
+        "substitutions": attr.string_dict(mandatory = True),
         "template": attr.label(
             mandatory = True,
             allow_single_file = True,
         ),
-        "substitutions": attr.string_dict(mandatory = True),
-        "data": attr.label_list(allow_files = True),
     },
     implementation = _sh_launcher_rule_impl,
     toolchains = [SH_TOOLCHAIN],
@@ -388,5 +389,91 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
     sh_launcher_rule(
         name = name,
         ament_setup = ament_setup,
+        **kwargs
+    )
+
+def _ros2_ament_cpp_library_impl(ctx):
+    src_output = ctx.actions.declare_file(ctx.attr.basename + ".cpp")
+    hdr_output = ctx.actions.declare_file(ctx.attr.basename + ".hpp")
+    ament_prefix_path = ctx.attr.ament_setup[Ros2AmentSetupInfo].ament_prefix_path or ""
+
+    substitutions = dicts.add(
+        ctx.attr.substitutions,
+        {
+            "{{ament_prefix_path}}": ament_prefix_path,
+            "{{header}}": hdr_output.short_path,
+        },
+    )
+
+    expand_template_impl(
+        ctx,
+        template = ctx.file.src_template,
+        output = src_output,
+        substitutions = substitutions,
+        is_executable = False,
+    )
+    expand_template_impl(
+        ctx,
+        template = ctx.file.hdr_template,
+        output = hdr_output,
+        substitutions = substitutions,
+        is_executable = False,
+    )
+
+    files = depset([src_output, hdr_output])
+    return [
+        DefaultInfo(
+            files = files,
+        ),
+    ]
+
+ros2_ament_cpp_library_src = rule(
+    attrs = {
+        "ament_setup": attr.label(
+            mandatory = True,
+            providers = [Ros2AmentSetupInfo],
+        ),
+        "basename": attr.string(mandatory = True),
+        "data": attr.label_list(allow_files = True),
+        "hdr_template": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "src_template": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "substitutions": attr.string_dict(mandatory = True),
+    },
+    implementation = _ros2_ament_cpp_library_impl,
+)
+
+def ros2_ament_setup_library(name, deps, idl_deps = None, **kwargs):
+    ament_setup = name + "_ament_setup"
+    testonly = kwargs.get("testonly", False)
+    ros2_ament_setup(
+        name = ament_setup,
+        deps = deps,
+        idl_deps = idl_deps,
+        package_name = native.package_name(),
+        tags = ["manual"],
+        testonly = testonly,
+    )
+    cpp_namespace = native.package_name().replace("/", "_") + "_" + name
+    ros2_ament_cpp_library_src(
+        name = name + "_srcs",
+        basename = name,
+        ament_setup = ament_setup,
+        src_template = "@com_github_mvukov_rules_ros2//ros2:ament_setup.cpp.tpl",
+        hdr_template = "@com_github_mvukov_rules_ros2//ros2:ament_setup.hpp.tpl",
+        substitutions = {"{{namespace}}": cpp_namespace},
+    )
+
+    cc_library(
+        name = name,
+        srcs = [":{}_srcs".format(name)],
+        deps = [],
+        data = [ament_setup],
+        testonly = testonly,
         **kwargs
     )
--- a/ros2/ament_setup.cpp.tpl
+++ b/ros2/ament_setup.cpp.tpl
@@ -2,7 +2,6 @@
 
 #include <cstdlib>
 #include <string>
-#include <iostream>
 
 namespace {{namespace}} {
 namespace {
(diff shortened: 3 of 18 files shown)


ros2/ament_setup.cpp.tpl
```diff
@@ -0,0 +1,39 @@
...
+    {
+      std::string ament_prefix_path_str = ament_prefix_path;
+      ament_prefix_path_str += ":";
+      ament_prefix_path_str += AMENT_PREFIX_PATH;
+      setenv(AMENT_PREFIX_PATH_ENV, ament_prefix_path_str.c_str(), 1);
+    }
+    else
+    {
+      setenv(AMENT_PREFIX_PATH_ENV, AMENT_PREFIX_PATH, 1);
+    }
```
mvukov, line 27: This is different compared to the logic in sh launcher. What's the reasoning here? Or a use-case?

ros2/test/pluginlib/ament_setup_library_test.cc
```diff
@@ -0,0 +1,49 @@
...
+  rcutils_logging_set_default_logger_level(RCUTILS_LOG_SEVERITY_DEBUG);
+  console_bridge::setLogLevel(console_bridge::CONSOLE_BRIDGE_LOG_DEBUG);
+
+  EXPECT_ANY_THROW(
+      pluginlib::ClassLoader<polygon_base::RegularPolygon> poly_loader(
+          "polygon_base", "polygon_base::RegularPolygon"));
+}
+
+TEST(TestAmentSetup, SetupAmentPrefixPath) {
+  ::ros2_test_pluginlib_ament_setup::setup_ament_prefix_path();
```
mvukov, line 37: Is it guaranteed that this init is definitely going to happen before the consumption of the prefix path? (which also happens during static var init phase)

mvukov (commented):
Hi, thanks for the contribution!

Out of curiosity, can you tell a bit more about the use-case where you don't want the sh launcher?

ros2/test/pluginlib/ament_setup_library_test.cc
woutervdstoel: You are correct that this function needs to be called very early. Are you aware of any initialization that happens prior to a ROS2 function being called from user code? In my use case of reading rosbags from an otherwise "pure" C++ library explicitly calling this function prior to calling a rosbags API function is sufficient.

woutervdstoel:
> Hi, thanks for the contribution!
> 
> Out of curiosity, can you tell a bit more about the use-case where you don't want the sh launcher?

I can't link the closed source code, but I implemented a C++ library that has to read from rosbags. The specific use case is a discrete-event simulator for deterministic resimulation of our robotics application.

Since this library is typically used from other code that isn't supposed to be ROS2-aware (also through python bindings), I really preferred it to be a "pure" `cc_library` and have it just work instead of requiring my (company-internal) users to use a sh wrapper. Executing tooling using `--run_under` is also much nicer when the application is "pure" C++.

ros2/test/pluginlib/ament_setup_library_test.cc
mvukov: I searched a bit, but couldn't find anything solid on this topic. There is no dependency in static variables in this case that could cause [Static Initialization Order Fiasco](https://en.cppreference.com/w/cpp/language/siof). Here we have to rely on ordering of init of static vars. TBH, I am surprised this works.

@ahans WDYT?

ros2/ament_setup.cpp.tpl
mvukov: As-is, this code allows other prefix paths to sneak in -- like from a regular ROS installation. That opens a can of worms and is a no-go for me at the moment.

I would expect one to use a single ros2_ament_setup_library per one binary. All deps given to ros2_ament_setup are run through machinery that will collect necessary info from all given deps and form an ament-compatible file tree that's going to be used by the app.

An alternative is to refactor your ros2_ament_setup_library to have have

```
        "ament_setups": attr.label_list(
            mandatory = True,
            providers = [Ros2AmentSetupInfo],
        ),
```
Then you can merge `DefaultInfo`s and form a unique prefix path e.g. `"path1:path2:...:pathn"` that's going to be baked in the .cc file. Keep in mind that in this case is expected there will be potentially massive amount of symlinks from different prefix path folders to same files. That's why I would go for option 1 and use one ros2_ament_setup_library per binary.

If you really need option 2 right now, please tell me more about your use-case.

Please let me know what you think.

woutervdstoel: > As-is, this code allows other prefix paths to sneak in -- like from a regular ROS installation. That opens a can of worms and is a no-go for me at the moment. I would expect one to use a single ros2_ament_setup_library per one binary. 

Could you explain the can of worms involved in doing this? Obviously sourcing regular ROS prior to running bazel-based applications is always a bad idea (I already found that  `@ros2cli` doesn't work correctly after sourcing ROS). The idea of this logic is that it's actually fine to combine the `AMENT_PREFIX_PATH` from multiple sources (e.g. one library that depends on rosbag plugins, another library that depends on some specific message types and so on) and we just keep expanding the `AMENT_PREFIX_PATH` as the relevant C++ code needs access to those types. Yes, you might end up with multiple instances of the same plugin libraries in the runfiles (each in different locations), but since the compilation options and source code is supposed to be the same within one bazel invocation, they should be equivalent and it doesn't matter which one is used.

The value I see in being able to do this, is that you can restrict the "ros-awareness" to specific cc_libraries. Each library just does what it needs to work correctly (this is a form of encapsulation). If you wanted to restrict the number of `ros2_ament_setup_library`s per binary, you will have to somehow use ros2-specific logic at the top level (i.e. the rule creating the executable) to confirm that. Or worse, not check it and silently run into hard-to-debug issues when you somehow do end up with multiple. In most cases such top-level logic is fine (ros2_cpp_binary actually works great!), but in my case I would really like to keep the ros-specific part encapsulated.

I can understand that this is not a super common use case, but other than the possible duplication of some symlinks in the runfiles I don't see a big problem with it.

woutervdstoel: > You just said there is a potential use-case where something might fail. :)

Is running with pre-sourced ROS something you want to support?

> > I can understand that this is not a super common use case, but other than the possible duplication of some symlinks in the runfiles I don't see a big problem with it.
> 
> OK, would my option 2 work for you?

If I understand your option 2 correctly, the problem I see with it is that you still need a single location where you bake the single merged path into the binary, which breaks the encapsulation.

mvukov: > Is running with pre-sourced ROS something you want to support?

This opens doors to various kind of problems, you already mentioned one. So, unless there is a general consensus among the users of the repo that we should do that, my answer is no.

ros2/test/pluginlib/ament_setup_library_test.cc
ahans: I think this is fine. `pluginlib` and `class_loader` maintain some global state, but that is only for already loaded libraries. If you try to load a plugin explicitly as in the present test or indirectly by requesting type support for some message type, eventually `ament_index_cpp` will be queried which does not cache anything, but always looks at the environment variable.

That being said, the shell script wrapper solution is certainly the cleaner one. It makes sure the environment variable is set from the start of the executable and doesn't change during runtime. This way it doesn't depend on a ROS implementation detail. If for some reason ament maintainers figure it's a good idea to cache some results and not read the environment variable each time (maybe tokenization of `AMENT_PREFIX_PATH` is considered too expensive...), the wrapper approach would still work, while the present one (changing the environment variable from within the executable) would break. That's not to say I'm against it. After all, before the shell script wrapper we had very similar functionality only available for Python targets and I argued at some point it would make sense to also have that for C++ targets. But one must be aware that there are certain caveats.

ros2/ament_setup.cpp.tpl
ahans: I totally agree that mixing stock ROS and `rules_ros2` is not a good idea.

@woutervdstoel The issue you saw with `ros2cli` was potentially due to the `PYTHONPATH` introduced by sourcing the stock ROS installation. Explicitly unsetting `PYTHONPATH` before running Bazel/`rules_ros2` targets fixed this for me.

As for the can of worms that appending to `AMENT_PREFIX_PATH` opens, I'm not sure. However, I would also argue for the approach where the final value for `AMENT_PREFIX_PATH` is assembled in one place, as that is more in line with the Bazel philosophy, where you give the build system as much information as possible and don't rely so much on assembling stuff at runtime. I agree that a single `ros2_ament_setup` per binary should be sufficient. By having a single ROS-aware top level rule that pulls all deps and takes care of ament setup you still have encapsulation. But you don't hide the fact that this somehow depends on ROS, which is actually a good thing IMHO. Too much magic also often can lead to hard-to-debug issues. The `ros2_ament_setup` also doesn't need to happen on the level of the binary. You can have a `cc_library` take care of that. If you really want to hide the fact that ROS is involved, you could also use custom versions of `cc_library` and `cc_{binary,test}` that transparently pass along the `Ros2AmentSetupInfo`s and only on the binary level create the actual `ros2_ament_setup` target. I'm not sure how much effort that would be, though.

The underlying issue here is that the natural way for a Bazel-based setup is very different from the native ROS way. Stock ROS relies heavily on resolving dependencies at runtime, while in a Bazel setup you want to decide as much as possible at build time. One way around this is creating package rules that produce packages in the native ROS format, including setup scripts. Over at [ApexAI/rules_ros](https://github.com/ApexAI/rules_ros/) that seems to be the idea. But with this approach you would lose some of the Bazel advantages such as being able to `bazel run` fine-grained targets. Instead, you would need one `ros2` target that depends on everything if you wanted to mimic the stock `ros2` behavior.

Overall, I think staying with a single `ros2_ament_setup` per binary would be more in line with the philosophy of `rules_ros2`. And I would trust my users can handle being exposed to the fact that ROS is used underneath. If you decide to go with the approach that involves more magic, make sure to cover that very well with tests. Otherwise that next ROS update might break something and leave you with that hard-to-debug issue. If you want encapsulation to be able to easily change ROS for something else some time in the future, I would say, forget it; that'd be a major refactor either way. ;-)

One last thought how one could work around the issue of some external `AMENT_PREFIX_PATH` sneaking in: You could maintain a "private" version (via another env variable or a static variable inside the executable) and use that to set `AMENT_PREFIX_PATH`, disregarding any value that `AMENT_PREFIX_PATH` had previously. However, since that would involve quite some unexpected "magic", I would most certainly not go for it...

ros2/test/pluginlib/ament_setup_library_test.cc
mvukov: OK, let's then proceed with the high-level idea of this PR. It should be well documented that the call to `setup_ament_prefix_path ` should happen "early enough" such that plugin registration works.

ros2/ament_setup.cpp.tpl
woutervdstoel: I fully agree with your statement that the native ROS2 plugin system does not play nicely with the bazel principle of specifying dependencies at build time. I also agree that trying to hide this fact using the approach from this PR is always going to result in a leaky abstraction. But I don't think that should be a reason not to allow it.

Forcing the usage of specific rules all the way to the top level of the dependency graph only to use a ROS library in some implementation detail is also a drawback, and I think that depending on the usage one negative can outweigh the other. I fully agree that in most cases it's better to use a specific top-level rule (or a single `ros2_ament_setup` per binary) but I still believe there's a use case for appending if one understands the risks.

This is why I just pushed a proposed change to make the default behavior to override the `AMENT_PREFIX_PATH` and the append-behavior opt-in. I also covered this with some unit tests.

It turns out the plugin loader only accepts one plugin providing types in a specific namespace so I also made the namespaces unique.

ros2/BUILD.bazel
```diff
@@ -21,6 +21,8 @@ exports_files([
     "test.bzl",
     "test.py.tpl",
     "topic.bzl",
+    "ament_setup.cpp.tpl",
+    "ament_setup.hpp.tpl",
```
mvukov, line 25: In this repo Google style guide is used. Hence, ament_setup.cc.tpl and ament_setup.h.tpl. Please alph-sort the list.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
         ament_setup = ament_setup,
         **kwargs
     )
+
+def _ros2_ament_cpp_library_impl(ctx):
+    src_output = ctx.actions.declare_file(ctx.attr.basename + ".cpp")
+    hdr_output = ctx.actions.declare_file(ctx.attr.basename + ".hpp")
```
mvukov, line 397: The good practice is to create a folder with the label name and put the files therein. Here:
```
src_output = ctx.actions.declare_file(ctx.attr.name + "/ament_setup.cc")
hdr_output = ctx.actions.declare_file(ctx.attr.name + "/ament_setup.h")
```

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+    ament_prefix_path = ctx.attr.ament_setup[Ros2AmentSetupInfo].ament_prefix_path or ""
+
+    substitutions = dicts.add(
+        ctx.attr.substitutions,
```
mvukov, line 401: Not used in the templates?

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+    ]
+
+ros2_ament_cpp_library_src = rule(
+    attrs = {
+        "ament_setup": attr.label(
+            mandatory = True,
+            providers = [Ros2AmentSetupInfo],
+        ),
+        "basename": attr.string(mandatory = True),
+        "data": attr.label_list(allow_files = True),
```
mvukov, line 437: Necessary?

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+        "hdr_template": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "src_template": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "substitutions": attr.string_dict(mandatory = True),
```
mvukov, line 446: Not used in the templates?

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+
+ros2_ament_cpp_library_src = rule(
+    attrs = {
+        "ament_setup": attr.label(
+            mandatory = True,
+            providers = [Ros2AmentSetupInfo],
+        ),
+        "basename": attr.string(mandatory = True),
+        "data": attr.label_list(allow_files = True),
+        "hdr_template": attr.label(
```
mvukov, line 438: -> _hdr_template set default to the header template.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "src_template": attr.label(
```
mvukov, line 442: -> _src_template, set the default to the src template.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+        ),
+    ]
+
+ros2_ament_cpp_library_src = rule(
+    attrs = {
+        "ament_setup": attr.label(
+            mandatory = True,
+            providers = [Ros2AmentSetupInfo],
+        ),
+        "basename": attr.string(mandatory = True),
```
mvukov, line 436: not needed.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+    testonly = kwargs.get("testonly", False)
+    ros2_ament_setup(
+        name = ament_setup,
+        deps = deps,
+        idl_deps = idl_deps,
+        package_name = native.package_name(),
+        tags = ["manual"],
+        testonly = testonly,
+    )
+    cpp_namespace = native.package_name().replace("/", "_") + "_" + name
```
mvukov, line 462: The namespace should be IMO `ctx.label.name`, then it's easy to document.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+        "src_template": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "substitutions": attr.string_dict(mandatory = True),
+    },
+    implementation = _ros2_ament_cpp_library_impl,
+)
+
+def ros2_ament_setup_library(name, deps, idl_deps = None, **kwargs):
```
mvukov, line 451: cpp_ament_setup_library is sufficient IMO.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+    )
+
+    files = depset([src_output, hdr_output])
+    return [
+        DefaultInfo(
+            files = files,
+        ),
+    ]
+
+ros2_ament_cpp_library_src = rule(
```
mvukov, line 430: cpp_ament_setup is sufficient. It's not really a library as it provides source files, only.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+    attrs = {
+        "ament_setup": attr.label(
```
mvukov, line 432: This can also be "ament_setups":  attr.label_list. Not necessarily for this PR, but you could make a TODO.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+    ros2_ament_setup(
+        name = ament_setup,
+        deps = deps,
+        idl_deps = idl_deps,
+        package_name = native.package_name(),
+        tags = ["manual"],
+        testonly = testonly,
+    )
+    cpp_namespace = native.package_name().replace("/", "_") + "_" + name
+    ros2_ament_cpp_library_src(
```
mvukov, line 463: Please tag this target as manual.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+        name = name + "_srcs",
```
mvukov, line 464: name + "_cpp_ament_setup" would be more consistent IMO.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+        basename = name,
+        ament_setup = ament_setup,
+        src_template = "@com_github_mvukov_rules_ros2//ros2:ament_setup.cpp.tpl",
+        hdr_template = "@com_github_mvukov_rules_ros2//ros2:ament_setup.hpp.tpl",
+        substitutions = {"{{namespace}}": cpp_namespace},
```
mvukov, line 469: Not needed, see comment above.

ros2/ament.bzl
```diff
@@ -390,3 +391,89 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+        tags = ["manual"],
+        testonly = testonly,
+    )
+    cpp_namespace = native.package_name().replace("/", "_") + "_" + name
+    ros2_ament_cpp_library_src(
+        name = name + "_srcs",
+        basename = name,
+        ament_setup = ament_setup,
+        src_template = "@com_github_mvukov_rules_ros2//ros2:ament_setup.cpp.tpl",
+        hdr_template = "@com_github_mvukov_rules_ros2//ros2:ament_setup.hpp.tpl",
```
mvukov, line 468: Can be hard-coded in the rule itself.

ros2/ament_setup.cpp.tpl
```diff
@@ -0,0 +1,30 @@
+#include "{{header}}"
```
mvukov, line 1: The header could simply be the `hdr_output.short_path`.

ros2/ament_setup.cpp.tpl
```diff
@@ -0,0 +1,30 @@
...
+
+#include <cstdlib>
+#include <string>
+
+namespace {{namespace}} {
```
mvukov, line 6: Should be the target name.

ros2/ament_setup.cpp.tpl
```diff
@@ -0,0 +1,30 @@
...
+namespace {
+
+constexpr const char* AMENT_PREFIX_PATH_ENV = "AMENT_PREFIX_PATH";
+constexpr const char* AMENT_PREFIX_PATH = "{{ament_prefix_path}}";
+
+}
+
+void setup_ament_prefix_path(bool append)
```
mvukov, line 14: Per the style guide  `SetUpAmentPrefixPath`. I propose `bool SetUpAmentPrefixPath(bool allow_appending)`, handle possible failures of setting up the env vars.

ros2/ament_setup.cpp.tpl
```diff
@@ -0,0 +1,30 @@
...
+{
+  const char* ament_prefix_path = std::getenv(AMENT_PREFIX_PATH_ENV);
+  if (ament_prefix_path && append)
```
mvukov, line 17: If allow_appending is false, the logic should be the same as in launch.sh.tpl.

ros2/ament_setup.hpp.tpl
```diff
@@ -0,0 +1,14 @@
...
+
+namespace {{namespace}} {
+
+/// Sets the AMENT_PREFIX_PATH to the install space of this package.
+///
+/// This function needs to be called prior to any attempt to use a plugin in this package.
+///
+/// @param append If true, the install space is appended to the existing AMENT_PREFIX_PATH.
+/// @note Using append=True can lead to unexpected behavior when another version of a
+/// plugin used in this package is already present in the AMENT_PREFIX_PATH.
```
mvukov, line 11: The documentation should be IMHO in the doc of the top-level Bazel macro. Folks most likely won't ever open this template. On the other hand, we'll eventually auto-generate docs for Bazel API.

ros2/test/pluginlib/BUILD.bazel
```diff
@@ -13,8 +14,8 @@ ros2_plugin(
     srcs = ["square.cc"],
     plugin_specs = [
         {
-            "class_type": "polygon_plugins::Square",
```
mvukov, line 16: I don't see how are those namespacing changes relevant to this PR. Please revert. We can discuss this in a separate PRs. Your tests should work with the existing plugins AFAICS.

ros2/test/pluginlib/BUILD.bazel
```diff
@@ -124,3 +125,29 @@ sh_test(
...
+
+ros2_ament_setup_library(
+    name = "square_ament_setup",
+    deps = [
+        ":square",
+    ],
+)
+
+cc_test(
+    name = "ament_setup_library_test",
```
mvukov, line 144: Please rename to ament_setup_library_tests and ament_setup_library_tests.cc, to be consistent with the rest of the repo.

ros2/test/pluginlib/ament_setup_library_test.cc
```diff
@@ -0,0 +1,88 @@
...
+  rcutils_logging_set_default_logger_level(RCUTILS_LOG_SEVERITY_DEBUG);
+  console_bridge::setLogLevel(console_bridge::CONSOLE_BRIDGE_LOG_DEBUG);
+
+  EXPECT_ANY_THROW(
+      pluginlib::ClassLoader<polygon_base::RegularPolygon> poly_loader(
+          "polygon_base", "polygon_base::RegularPolygon"));
+}
+
+TEST(TestAmentSetup, SetupSingleAmentPrefixPath) {
+  ::ros2_test_pluginlib_triangle_ament_setup::setup_ament_prefix_path();
```
mvukov, line 38: To be on the safe side, please create a fixture that in TearDown override unsets AMENT_PREFIX_PATH. I'd suggest that in the static setup function of the fixture you set up logging to reduce code dup.

ros2/test/pluginlib/ament_setup_library_test.cc
```diff
@@ -0,0 +1,88 @@
...
+  // this will reset the AMENT_PREFIX_PATH to only contain the triangle library.
+  ::ros2_test_pluginlib_triangle_ament_setup::setup_ament_prefix_path();
+  {
+    pluginlib::ClassLoader<polygon_base::RegularPolygon> poly_loader(
+        "polygon_base", "polygon_base::RegularPolygon");
+    EXPECT_ANY_THROW(poly_loader.createSharedInstance("square_plugin::Square"));
+  }
+
+  // and now only the square library
+  ::ros2_test_pluginlib_square_ament_setup::setup_ament_prefix_path();
```
mvukov, line 82: Do you anticipate users would do such shenanigans in their code? Curious. :)

mvukov (changes requested):
I think the general direction is OK, I left comments on how to improve this PR further.

ros2/test/pluginlib/ament_setup_library_test.cc
woutervdstoel: I sure hope not :) It's more of a test of robustness to make sure the class loader doesn't cache plugin libraries statically between calls (which wouldn't just invalidate the concept of appending but also of overriding.

ros2/ament.bzl
woutervdstoel: I'm using the name of the wrapper target (basename) to avoid the `_srcs` suffix of the implementation target leaking into the generated code.

ros2/ament.bzl
woutervdstoel: `expand_template_impl` looks at this (I haven't dug into why).

ros2/ament.bzl
woutervdstoel: Done, but I could see one downside to this in the "multiple `cpp_ament_setup` targets with appending" case: if two targets have the same name in different package we will violate the ODR. I don't think this is a big deal.

ros2/ament.bzl
woutervdstoel: What use case do you see for this? I'm not sure I'm following.

ros2/ament.bzl
woutervdstoel: This target generates source files for the target with name `name`. I think that makes `_srcs` appropriate?

ros2/ament_setup.hpp.tpl
woutervdstoel: Added additional docs there - I don't think it hurts to repeat the warning in both places.

ros2/test/pluginlib/BUILD.bazel
woutervdstoel: I'm still looking for the bit of code in the pluginlib, classloader and ament code that makes these changes required. Somewhere in that code an assumption seems to be made that all classes in a single namespace are exported by the same plugin. The code doesn't work without these changes. I'm happy to extract them into another PR as well if you prefer.

mvukov: Why would we need those changes in this PR except for testing? Can you refactor the tests to work with the current test code?

ros2/ament.bzl
mvukov: It could be useful, nothing to be done in this PR.

ros2/ament_setup.h.tpl
```diff
@@ -0,0 +1,14 @@
...
+
+namespace {{namespace}} {
+
+/// Sets the AMENT_PREFIX_PATH to the install space of this package.
+///
+/// This function needs to be called prior to any attempt to use a plugin in this package.
+///
+/// @param allow_appending If true, the install space is appended to the existing AMENT_PREFIX_PATH.
+/// @note Using append=True can lead to unexpected behavior when another version of a
+/// plugin used in this package is already present in the AMENT_PREFIX_PATH.
```
mvukov, line 11: It's doubling the potential maintenance. Therefore, I'd prefer docs to only live in the Bazel macro docs.

ros2/test/pluginlib/ament_setup_library_tests.cc
```diff
@@ -0,0 +1,95 @@
...
+#include "ros2/test/pluginlib/regular_polygon.h"
+#include "ros2/test/pluginlib/square_ament_setup/ament_setup.h"
+#include "ros2/test/pluginlib/triangle_ament_setup/ament_setup.h"
+
+using ::testing::DoubleNear;
+using ::testing::Eq;
+
+class TestAmentSetup : public ::testing::Test {
+ public:
+  void SetUp() override {
```
mvukov, line 30: You can just use http://google.github.io/googletest/reference/testing.html#Test::SetUpTestSuite.

ros2/ament.bzl
```diff
@@ -390,3 +391,99 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+)
+
+def cpp_ament_setup_library(name, deps, idl_deps = None, **kwargs):
+    """ Generates a C++ library that contains the ament setup code for a ROS2 package.
+
+    This can be useful in contexts where using a wrapper script (as is the case with the regular ros2_xxx rules)
+    is not desirable. The generated function can be called with an optional allow_append argument, which causes the
+    function to append to the AMENT_PREFIX_PATH instead of overwriting it. Beware that when using this mode when
+    different versions of the same plugin exist, the behavior might be unexpected.
+
```
mvukov, line 457: Please add docs for usage of the generated header:
```cpp
#include "<package_name>/<name>/ament_setup.h"
// then call
::<name>::SetUpAmentPrefixPath();
```

ros2/ament.bzl
mvukov: OK, let's leave it as-is :+1: .

ros2/ament.bzl
```diff
@@ -390,3 +391,99 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+    implementation = _cpp_ament_setup_impl,
+)
+
+def cpp_ament_setup_library(name, deps, idl_deps = None, **kwargs):
+    """ Generates a C++ library that contains the ament setup code for a ROS2 package.
+
+    This can be useful in contexts where using a wrapper script (as is the case with the regular ros2_xxx rules)
+    is not desirable. The generated function can be called with an optional allow_append argument, which causes the
+    function to append to the AMENT_PREFIX_PATH instead of overwriting it. Beware that when using this mode when
+    different versions of the same plugin exist, the behavior might be unexpected.
```
mvukov, line 456: Please add a sentence that the generated cpp function should be called "early enough", where early enough is most likely to be before any ament resource is used.

ros2/ament.bzl
```diff
@@ -390,3 +391,99 @@ def sh_launcher(name, deps, idl_deps = None, **kwargs):
...
+        ),
+    ]
+
+cpp_ament_setup = rule(
+    attrs = {
+        "ament_setup": attr.label(
+            mandatory = True,
+            providers = [Ros2AmentSetupInfo],
+        ),
+        "basename": attr.string(
```
mvukov, line 434: OK, we can stick with basename. It doesn't need to be mandatory IMO. You can do something like in the impl:
```py
basename = ctx.attr.basename or ctx.attr.name
src_output = ctx.actions.declare_file(basename + "/ament_setup.cc")
...
```

ros2/test/pluginlib/BUILD.bazel
mvukov: OK, fair enough; it's strange this doesn't work btw (have no idea whether this is supposed to work w/ regular ROS workspaces). Anyway, you can create another set of simple test plugins, e.g. use a genrule to generate source files and plugin targets using list comprehension in the build file. The existing tests and the new ones do not test the same functionality, they are unrelated IMO.

woutervdstoel: I looked a bit further and the reason it wouldn't work is that the generated package name of both the `triangle` and `square` plugins are the same (because of the way `_get_package_name` is implemented in `ament.bzl`. I guess in the normal colcon ros2 world these are manually specified by the user. I don't think we could change this logic without breaking existing code, but at least now we know why it wouldn't work before. I will create some alternative tests for my PR instead as you suggested.

mvukov (approved):
Thanks for all your efforts :) LGTM!

ros2/test/pluginlib/BUILD.bazel
mvukov: It might be that indeed that use-case is not covered there. We can refactor/upgrade the relevant code when needed.

PR #116, opened 2023-05-26
Fix gtest fetch issue

fernride-jissing:
bazel fetch errors can also cause issues with bazel queries and bazel fetch can detect dependency changes if run in CI.

### DIFF
--- a/repositories/ros2_repositories_impl.bzl
+++ b/repositories/ros2_repositories_impl.bzl
@@ -62,7 +62,7 @@ def ros2_repositories_impl():
     maybe(
         http_archive,
         name = "iceoryx",
-        strip_prefix = "iceoryx-2.0.2",
+        strip_prefix = "iceoryx-2.0.3",
         build_file = "@com_github_mvukov_rules_ros2//repositories:iceoryx.BUILD.bazel",
         sha256 = "8f391696daf2e63da9437aab8d7154371df630fc93876479f2e84c693fc1ba5a",
         url = "https://github.com/eclipse-iceoryx/iceoryx/archive/refs/tags/v2.0.3.tar.gz",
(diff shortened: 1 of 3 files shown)


repositories/ros2_repositories_impl.bzl
```diff
@@ -62,7 +62,7 @@ def ros2_repositories_impl():
     maybe(
         http_archive,
         name = "iceoryx",
-        strip_prefix = "iceoryx-2.0.2",
+        strip_prefix = "iceoryx-2.0.3",
```
mvukov, line 65: This is seems to be unrelated to this PR.

mvukov (commented):
TBH, I don't understand what doesn't work for you that requires this gtest patch. Can you elaborate, please?

repositories/ros2_repositories_impl.bzl
mvukov: Fixed in https://github.com/mvukov/rules_ros2/pull/118.

fernride-jissing:
> TBH, I don't understand what doesn't work for you that requires this gtest patch. Can you elaborate, please?

Sorry for the late reponse. It's an issue that's simple to reproduce:

```bash
git clone git@github.com:mvukov/rules_ros2.git
cd rules_ros2
bazel fetch //...
```

It will also cause issues if one want so run `bazel query` that includes external dependencies.

mvukov:
OK, thanks for the explanation. I'd propose https://github.com/mvukov/rules_ros2/pull/120 as an alternative. I consider patches as a last resort, typically. Please let me know what you think.

fernride-jissing:
> OK, thanks for the explanation. I'd propose #120 as an alternative. I consider patches as a last resort, typically. Please let me know what you think.

Looks good to me! Thanks :)

PR #113, opened 2023-05-18
Integrate ros-foxglove-bridge

mvukov:

PR #117, opened 2023-05-28
Fix/rclcpp ts libs ownership

mvukov:
This PR fixes an issue discovered within https://github.com/mvukov/rules_ros2/pull/113. Turned out that type-support shared libraries are deleted (deallocated) too soon when one deletes e.g. a publisher (happens in another PR). As a result of that, apps segfault.

In the provided, new, rclcpp path I took a different route now: the node holds now a ts_libs shared ptrs as well, ensuring that the ts_libs live long enough (well, until the node is destroyed). This introduces extra deps in ROS 2 Node class, but IMO changes are fairly small. I admit this is a yet another band-aid until I can allocate some time to refactor how type support libs are built by Bazel.

@ahans PTAL.


mvukov:
BTW, I tested this with #113 and this PR helps -- no more segfaults when a published/client is removed.

ahans (approved):
LGTM!

I'm not sure why this fixes the segfault in #113, probably the shared library gets unloaded a little bit later now, but it's essentially very similar to the patches we had before. Maybe some Foxglove code relies on that shared library that's never unloaded, just like the publisher/subscription base classes do. Anyway, if this fixes the issue, great!

PR #118, opened 2023-05-31
Fix iceoryx compilation after version update

LiyouZhou:


mvukov (approved):
LGTM! Thanks for fixing this.

Issue #144, opened 2023-07-24
Issue with cross compiling rnw_cyclonedds towards aarch64


henriksod:
Hi, I am using [gcc_toolchain](https://github.com/aspect-build/gcc-toolchain) to cross compile ros2 nodes towards aarch64 using this repo. I am having an issue with `@ros2_rmw_cyclonedds//:rmw_cyclonedds`.

**How to reproduce:**
-----

`BAZEL_VERSION=6.2.1`

`rules_ros2: 89dd5fa0add476e85a438c8575f353ecf6162c57`

<details>
<summary>WORKSPACE</summary>

```bazel
# WORKSPACE

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

# Toolchain: aarch64-linux-gnueabihf
http_archive(
    name = "aspect_gcc_toolchain",
    sha256 = "8850373f24d3f8bb6e8f36e3e8e7edc93d948964f8f201e920af2c8ffba2002c",
    strip_prefix = "gcc-toolchain-4bd1f94536ee92b7c49673931773038d923ee86e",
    url = "https://github.com/aspect-build/gcc-toolchain/archive/4bd1f94536ee92b7c49673931773038d923ee86e.tar.gz",
)

load("@aspect_gcc_toolchain//toolchain:repositories.bzl", "gcc_toolchain_dependencies")

gcc_toolchain_dependencies()

load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")

bazel_skylib_workspace()

load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")

aspect_bazel_lib_dependencies()

load("@aspect_gcc_toolchain//toolchain:defs.bzl", "ARCHS", "gcc_register_toolchain")

# Register aarch64 toolchain
gcc_register_toolchain(
    name = "gcc_toolchain_aarch64",
    target_arch = ARCHS.aarch64,
)

http_archive(
    name = "com_github_mvukov_rules_ros2",
    sha256 = "c1ff135dd1a6a5c518357285611b1c4de4af6eb9249bf007a21479e35b1a6006",
    strip_prefix = "rules_ros2-89dd5fa0add476e85a438c8575f353ecf6162c57",
    url = "https://github.com/mvukov/rules_ros2/archive/89dd5fa0add476e85a438c8575f353ecf6162c57.tar.gz",
)

[...]

PR #148, opened 2023-07-29
Add fastrtps (Fast-DDS) as supported dds middleware

henriksod:
As requested in this discussion https://github.com/mvukov/rules_ros2/discussions/145 

This is an attempt to add Fast-DDS as a supported dds middleware to this repo. I am using versions compatible with ROS2 Humble.

Added repo dependencies:
- foonathan_memory
- fastcdr
- fastrtps
- rmw_fastrtps

I have modified `rmw_implementation.BUILD.bazel` to allow for a build flag `dds_vendor`, which will be `cyclonedds` by default.
This is how you can set it to use fastdds instead:
```bazel
# To use FastDDS, invoke Bazel with `--config=fastdds`.
build:fastdds --@ros2_rmw_implementation//:dds_vendor=fastdds
```

Tests are currently failing, need help to figure out what might be causing the error:

<details>
<summary>test.log</summary>

```bash
Executing tests from //ros2/test:generic_publisher_tests
-----------------------------------------------------------------------------
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from TestGenericPublisher
[ RUN      ] TestGenericPublisher.WhenPublisherDeletedInThread_EnsureCleanNodeCleanup

>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

  'error not set, at external/ros2_rcl/rcl/src/rcl/node.c:263'

with this new error message:

  'rcl node's rmw handle is invalid, at external/ros2_rcl/rcl/src/rcl/node.c:415'

rcutils_reset_error() should be called after error handling to avoid this.
<<<
[ERROR] [1690534755.799629788] [rcl]: Failed to fini publisher for node: 1
unknown file: Failure
C++ exception with description "failed to initialize rcl node: rcl node's rmw handle is invalid, at external/ros2_rcl/rcl/src/rcl/node.c:415" thrown in the test body.
[  FAILED  ] TestGenericPublisher.WhenPublisherDeletedInThread_EnsureCleanNodeCleanup (14 ms)
[----------] 1 test from TestGenericPublisher (14 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (14 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestGenericPublisher.WhenPublisherDeletedInThread_EnsureCleanNodeCleanup

 1 FAILED TES
```
</details>

PR #151, opened 2023-07-29
Support cross compilation to aarch64

henriksod:
Solving the issue discussed here: https://github.com/mvukov/rules_ros2/issues/144

I also implemented a feature to cross compile towards aarch64
See https://github.com/mvukov/rules_ros2/discussions/150 for more info.

- Patch cyclonedds to remove -flto
- Added gcc_toolchain and openssl cross compile support for aarch64

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -46,3 +46,31 @@ load(
 )
 
 install_rules_ros2_resolver_deps()
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "aspect_gcc_toolchain",
+    sha256 = "8850373f24d3f8bb6e8f36e3e8e7edc93d948964f8f201e920af2c8ffba2002c",
+    strip_prefix = "gcc-toolchain-4bd1f94536ee92b7c49673931773038d923ee86e",
+    url = "https://github.com/aspect-build/gcc-toolchain/archive/4bd1f94536ee92b7c49673931773038d923ee86e.tar.gz",
+)
+
+load("@aspect_gcc_toolchain//toolchain:repositories.bzl", "gcc_toolchain_dependencies")
+
+gcc_toolchain_dependencies()
+
+load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
+
+bazel_skylib_workspace()
+
+load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")
+
+aspect_bazel_lib_dependencies()
+
+load("@aspect_gcc_toolchain//toolchain:defs.bzl", "ARCHS", "gcc_register_toolchain")
+
+gcc_register_toolchain(
+    name = "gcc_toolchain_aarch64",
+    target_arch = ARCHS.aarch64,
+)
--- a/.bazelrc
+++ b/.bazelrc
@@ -30,6 +30,10 @@ build:clang --repo_env=CC=clang
 build:clang --repo_env=CXX=clang++
 build:clang --linkopt="-fuse-ld=lld"
 
+# To cross-compile towards aarch64, invoke Bazel with `--config=aarch64_linux_gcc`.
+build:platform_build --incompatible_enable_cc_toolchain_resolution
+build:aarch64_linux_gcc --config=platform_build --platforms=@aspect_gcc_toolchain//platforms:aarch64_linux
+
 # Load any settings specific to the current user.
 # user.bazelrc should appear in .gitignore so that settings are not shared with
 # team members. This needs to be last statement in this config,
--- a/repositories/openssl.BUILD.bazel
+++ b/repositories/openssl.BUILD.bazel
@@ -1,4 +1,39 @@
+load("@bazel_skylib//lib:selects.bzl", "selects")
 load("@rules_foreign_cc//foreign_cc:defs.bzl", "configure_make")
+
+# support cross-compilation
+# otherwise openssl will configure towards the host system
+selects.config_setting_group(
+    name = "linux_aarch64",
+    match_all = [
+        "@platforms//os:linux",
+        "@platforms//cpu:aarch64",
+    ],
+)
+
+selects.config_setting_group(
+    name = "linux_x86_64",
+    match_all = [
+        "@platforms//os:linux",
+        "@platforms//cpu:x86_64",
+    ],
+)
+
+selects.config_setting_group(
+    name = "macos_x86_64",
+    match_all = [
+        "@platforms//os:macos",
+        "@platforms//cpu:x86_64",
+    ],
+)
+
+selects.config_setting_group(
+    name = "macos_aarch64",
+    match_all = [
+        "@platforms//os:macos",
+        "@platforms//cpu:aarch64",
+    ],
+)
 
 filegroup(
     name = "all_srcs",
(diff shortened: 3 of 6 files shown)


WORKSPACE
```diff
@@ -46,3 +46,31 @@ load(
 )
 
 install_rules_ros2_resolver_deps()
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
```
ahans, line 52: This is nothing that we strictly need in the global `WORKSPACE` file, as it is only used to test cross compilation. I'd prefer adding that to the `examples` workspace or even create a new `WORKSPACE` for this.

@mvukov What do you think?

henriksod: Hi! I suggest it is run through CI anyways, to ensure compilation towards aarch64 doesn't break.

ahans: Yes, I'm all for running it in CI, just setup a bit differently. We already have to jobs, one for the main workspace and one for examples. Would be easy to add a third. To trigger aarch64 compilation you need an extra job anyway.

mvukov:
Before I started working on this repo, I was working on rules_ros. And there a holy goal for me was to be able to easily cross-compile a deployment (launch specs + nodes). Cross-compiling of the C++ code was fairly easy, but, then there was Python code. Back then I was mostly interested in C++ nodes (pretty much now as well), thus, on the Python side only for launch mechanisms. The problem was that some of the Python deps had compiled extensions. Python ecosystem is wild and I didn't want to invest much of my time figuring out all small details. TL;DR I got rid of some packages I didn't need in roslaunch and found python-only alternatives to replace packages with compiled extensions. Still hacky here and there, but this worked. (I didn't even bother configuring appropriate Python toolchain back then...)

In the meantime, arm64 dev stations became a commodity, e.g. https://www.ipi.wiki/products/ampere-altra-developer-platform -- at least for a company setting. Therefore, I am questioning cross-compilation these days. For hobbyists it still makes sense, naturally :)

On a high level: yay :100: for this PR, but we should also think about how far we want to go with this. Do you want to work on figuring out how to make Python code working on arm64 as well? I am fine if we scope this that we're interested at the moment in Python packages only necessary for launching nodes: e.g. that top-level ros2_launch targets can be correctly cross-compiled and packaged for exec on an arm64 platform.

For this to work, we'll need a transitive closure for arm64 made from requirements.txt. pip-compile cannot resolve packages for another platform, btw. There exists something like https://pypi.org/project/pip-compile-cross-platform/, that I think uses poetry package under the hood. Then, let's say we have requirements_lock_amd64.txt and requirements_lock_arm64.txt. We'd need to figure out a convenient way to consume pip deps w/o forcing users to copy/paste selects in the build files.

Regarding CI. I think we want to have a workflow where we cross-compile e.g. a test target on a amd64 host for arm64 target and then in another job run that test on arm64 host to see whether it actually works. We can't do that with github actions at the moment, but we could do that on circleci. At the moment I think that would be an overkill to do for every PR, but once a week sounds good.

Regarding the lto patch: would this be necessary if we would natively compile code on an arm64 platform? I suspect not. If we wouldn't need that for native compilation, then it should be possible to tweak cross-compilation.

henriksod:
> On a high level: yay 💯 for this PR, but we should also think about how far we want to go with this. Do you want to work on figuring out how to make Python code working on arm64 as well? I am fine if we scope this that we're interested at the moment in Python packages only necessary for launching nodes: e.g. that top-level ros2_launch targets can be correctly cross-compiled and packaged for exec on an arm64 platform.

Hi! I am actually able to cross compile ros2_launch targets completely fine towards aarch64. Maybe there have been updates to rules_python since last time. Currently I have other struggles that I need to sort out. Seems like [micro-ros-agent](https://github.com/micro-ROS/micro-ROS-Agent) requires `rmw_fastrtps_cpp` and does not support `rmw_fastrtps_dynamic_cpp` as is currently implemented in https://github.com/mvukov/rules_ros2/pull/148. I need to either add support for that, rewrite the micro-ros-agent entirely to support introspection, or switch to Zenoh. 

```bash
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_catkin_pkg/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_docutils/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_empy/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_lark_parser/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_netifaces/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_numpy/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_packaging/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_pyparsing/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_python_dateutil/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_pyyaml/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_setuptools/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_six/
drwxr-xr-x  3 pi pi 4096 Jan  1  2000 rules_ros2_pip_deps_types_pkg_resources/
drwxr-xr-x  6 pi pi 4096 Jan  1  2000 rules_ros2_python_aarch64-unknown-linux-gnu/
```

```bash
pi@balancingrobot:/opt/balancing_robot $ ./launch_package
[INFO] [launch]: All log files can be found below /home/pi/.ros/log/2023-08-01-23-22-55-133995-balancingrobot-139566
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [micro_ros_agent-1]: process started with pid [139570]
[INFO] [micro_ros_listener_example_task-2]: process started with pid [139572]
[micro_ros_agent-1] [1690924975.863841] info     | TermiosAgentLinux.cpp | init                     | running...             | fd: 3
[micro_ros_agent-1] [1690924975.865139] info     | Root.cpp           | set_verbose_level        | logger setup           | verbose_level: 4
[micro_ros_listener_example_task-2] [INFO] [1690924976.579878127] [micro_ros_listener_example_task]: I am alive
```

mvukov:
What does the output of `ls rules_ros2_pip_deps_netifaces/site-packages` look like? Do you get arm64 .so binary out of the box?

mvukov:
> Seems like [micro-ros-agent](https://github.com/micro-ROS/micro-ROS-Agent) requires `rmw_fastrtps_cpp` and does not support `rmw_fastrtps_dynamic_cpp` as is currently implemented in #148. I need to either add support for that, rewrite the micro-ros-agent entirely to support introspection, or switch to Zenoh.

IMO, it still makes sense to finish up #148 since you already made quite some progress. Off the top of my head, a rewrite of micro-ros-agent sounds more complicated than supporting rmw_fastrtps_cpp in this repo. :)

mvukov:
So, I'm not sure Python stuff would just work out of the box. That's why I asked :arrow_up: 

> What does the output of ls rules_ros2_pip_deps_netifaces/site-packages look like? Do you get arm64 .so binary out of the box?

Also, I had a question:

> Regarding the lto patch: would this be necessary if we would natively compile code on an arm64 platform? I suspect not. If we wouldn't need that for native compilation, then it should be possible to tweak cross-compilation.

Please reply when you find some time.

aspect_gcc_toolchain has some huge deps (e.g. a rootfs) and I would rather set up an example in a separate workspace. Next to that, would be great if we can test whether the cross-compilation works. I made a comment that we could try to set this up on circleci, eventually.

ahans:
> Regarding the lto patch: would this be necessary if we would natively compile code on an arm64 platform? I suspect not. If we wouldn't need that for native compilation, then it should be possible to tweak cross-compilation.

I have tried this out a couple weeks ago. It did work just fine. So the `-flto -fPIC` incompatibility seems to only apply to cross compilation. Most likely explanation to me is still a GCC bug.

At any rate, IMHO this issue is only remotely related to `rules_ros2`, so a FAQ entry to document that this issue is known and how to work around it would be good enough. An extra workspace that shows the workaround in detail (patching CycloneDDS's `CMakeLists.txt` or selecting `RelWithDebInfo`) would of course be even nicer than just the FAQ entry.

mvukov:
Hm, does this looks like a gcc bug at a specific gcc version?

PR #169, opened 2023-08-22
Use `exec` in launch wrapper

ahans:
This adds `exec` to the call of the actual binary in the launch script template. This way, no new process is created for the actual binary, but it takes the place of the wrapper script (including its PID). Without this, ROS 2 launch cannot properly communicate with the binary, as it sends signals to the wrapper script, which by default doesn't forward signals to a child process (the binary in this case).


mvukov (approved):
:100:

PR #173, opened 2023-08-31
Bzlmod Migration

silicon-ninja:
Hey @mvukov , 

Will keep on updating the PR ... as we move on `bzlmod` transition


silicon-ninja:
So few things !

- I've added `.devcontainer` for vscode development its easier to develop inside docker for this project to isolate all the code.

- Trying to make sure that to see if poetry would help us cross compile the code to `aarch64` only issue i found was `netifaces` maybe has to be recompiled with clang for the wheel file since it only supports `x86` and poetry throws an error since it could not find it for my laptop (ARM64)

- I am also expermenting with [hermetic_cc_toolchain](https://github.com/uber/hermetic_cc_toolchain) which uses `zig` to compile the c/cpp code ! lot faster and less configuration 

Please let me know your thoughts on this ! so we could proceed further accordingly ... Thanks 😁

mvukov:
This work is still fresh, so converted it to draft. As-is right now, the PR has some unrelated changes. If you want to introduce bzlmod, I'd suggest to "just" do that in this PR and leave rules_poetry, cross-compilation, dev containers for follow-up PRs. This way the PR is going to be much shorter and easier to review.

MODULE.bazel
```diff
@@ -0,0 +1,63 @@
+"""Module Declaration for rules_ros2."""
+
+module(
+    name = "rules_ros2",
```
mvukov, line 4: Anyone can make a workspace/module rules_ros2. I gave kind-of unique name com_github_mvukov_rules_ros2 to distinguish this work from possible other attempts to hook ROS 2 to Bazel. Thus, please stick with com_github_mvukov_rules_ros2. If it turns out rules_ros2 would be sufficient/desired, we can always have a follow-up PR just to rename stuff. Let's have this PR as simple and minimal as possible.

oh-tarnished:
> This work is still fresh, so converted it to draft. As-is right now, the PR has some unrelated changes. If you want to introduce bzlmod, I'd suggest to "just" do that in this PR and leave rules_poetry, cross-compilation, dev containers for follow-up PRs. This way the PR is going to be much shorter and easier to review. 

Hey thank you, So well I could link multiple issues with this PR for rules_poetry, cross-compilation, dev containers to follow up .... I thought to wrap it up in one go. but sure if you wanna do that I could only make it for `bzlmod`

MODULE.bazel
oh-tarnished: Yeah I understand that, but you can alias your `MODULE.bazel` to this, 

```bazel
bazel_dep ( name = "rules_ros2" , version = "0.1.0" , repo_name = "com_github_mvukov_rules_ros2 ") 
```

it would be convenient for a user to call in `BUILD.bazel` files is what I thought but sure.

mvukov: Since we're going to continue to use the workspace-based workflow for a while, I'd prefer the name to be the same; hence: com_github_mvukov_rules_ros2.

mvukov:
Follow-ups should go to follow-up PRs, this one should be as small as possible.

MODULE.bazel
silicon-ninja: sure will keep the module name the same

jondo2010:
Hi folks, I've also been actively working on migrating our workspaces to bzlmod, including `rules_ros2`.

As a first step, I migrated `rules_pcl`, PR is here: https://github.com/kgreenek/rules_pcl/pull/17

PR #190, opened 2023-10-11
Add cpp_tf2_geometry_msgs cpp library

kgreenek:


repositories/geometry2.BUILD.bazel
```diff
@@ -158,6 +158,19 @@ ros2_cpp_binary(
     ],
 )
 
+ros2_cpp_library(
+    name = "cpp_tf2_geometry_msgs",
+    hdrs = ["tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.hpp"],
```
mvukov, line 163: Should we also include the corresponding .h file, just in case someone would still use it?

kgreenek: I'll leave that up to you. I left it out because it's long deprecated and other similar libraries in rules_ros2 didn't include the .h file.

mvukov: OK, let's leave it as-is and add if necessary.

mvukov (approved):
Thanks for your contribution.

PR #196, opened 2023-10-14
Remove recorder from diagnostics test

### LINKED ISSUE #195: Fix flaky //ros2/test/diagnostics:diagnostic_heartbeat_test

ahans:
We robustify this test by not relying on a bag recorder, but instead have a subscriber within the same process as the test. Since we have the messages available without extra effort, we use the chance to also check the contents of the messages.

Fixes #195.

### DIFF
--- a/ros2/test/diagnostics/BUILD.bazel
+++ b/ros2/test/diagnostics/BUILD.bazel
@@ -1,7 +1,5 @@
-load("@com_github_mvukov_rules_ros2//ros2:bag.bzl", "ros2_bag")
 load("@com_github_mvukov_rules_ros2//ros2:cc_defs.bzl", "ros2_cpp_binary")
 load("@com_github_mvukov_rules_ros2//ros2:test.bzl", "ros2_test")
-load("@rules_ros2_pip_deps//:requirements.bzl", "requirement")
 
 ros2_cpp_binary(
     name = "diagnostic_publisher",
@@ -12,27 +10,19 @@ ros2_cpp_binary(
     ],
 )
 
-ros2_bag(
-    name = "bag",
-    idl_deps = [
-        "@ros2_common_interfaces//:diagnostic_msgs",
-        "@ros2_rcl_interfaces//:rcl_interfaces",
-        "@ros2_rosbag2//:rosbag2_interfaces",
-    ],
-)
-
 ros2_test(
     name = "diagnostic_heartbeat_test",
     size = "small",
     data = [
         "aggregator_config.yaml",
-        ":bag",
     ],
     launch_file = "tests.py",
     nodes = [
         ":diagnostic_publisher",
         "@ros2_diagnostics//:aggregator_node",
     ],
-    tags = ["cpu:2"],
-    deps = [requirement("pyyaml")],
+    deps = [
+        "@ros2_common_interfaces//:py_diagnostic_msgs",
+        "@ros2_rclpy//:rclpy",
+    ],
 )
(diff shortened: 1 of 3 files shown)


ros2/test/diagnostics/BUILD.bazel
```diff
@@ -12,27 +11,20 @@ ros2_cpp_binary(
...
     data = [
         "aggregator_config.yaml",
-        ":bag",
     ],
     launch_file = "tests.py",
     nodes = [
         ":diagnostic_publisher",
         "@ros2_diagnostics//:aggregator_node",
     ],
-    tags = ["cpu:2"],
```
ahans, line 36: Not sure what this was good for. Can revert if this does anything useful.

ros2/test/diagnostics/diagnostic_publisher.cc
```diff
@@ -25,21 +24,17 @@ class Publisher : public rclcpp::Node {
     diagnostic_updater_.setHardwareID("none");
     diagnostic_updater_.add("", &diagnostic_heartbeat_,
                             &diagnostic_updater::Heartbeat::run);
-
-    auto timer_callback{[this]() {
-      // do nothing
-    }};
-    timer_ = create_wall_timer(std::chrono::milliseconds(20), timer_callback);
```
ahans, line 32: This seemed unnecessary.

ros2/test/diagnostics/tests.py
```diff
@@ -11,18 +11,18 @@
...
 import launch.actions
 import launch_ros.actions
 import launch_testing.actions
 import launch_testing.asserts
 import launch_testing.markers
-import yaml
-
-bag_dir = os.path.join(os.environ['TEST_TMPDIR'], 'bag')
+import rclpy
+from diagnostic_msgs.msg import DiagnosticArray
```
mvukov, line 24: Import `diagnostic_msgs.msg`.

ros2/test/diagnostics/tests.py
```diff
@@ -11,18 +11,18 @@
...
+from rclpy import node
```
mvukov, line 25: IMO, `node` is too generic so `import rclpy.node` might be better.

ros2/test/diagnostics/BUILD.bazel
mvukov: https://bazel.build/reference/test-encyclopedia: this tag give tests more cores. I put this for robustness, won't harm to have this IMO.

mvukov (commented):
Very  nice, much appreciated. I left a couple of minor comments.

ros2/test/diagnostics/tests.py
ahans: Went with a single `import rclpy` now.

ros2/test/diagnostics/BUILD.bazel
ahans: At the same time, we don't _need_ it any more. With a timeout of 10 s, a single core is easily able to handle this. I ran it with this on my laptop with 4 physical cores and it passed just fine:
```
bazel test --runs_per_test=100 --jobs=100 --local_cpu_resources=100 //ros2/test/diagnostics:diagnostic_heartbeat_test
```
So would keep it remove, since it adds unnecessary noise. If you insist, of course, I will put it back.

ahans:
Addressed comments. Please have another look!

mvukov (approved):

PR #231, opened 2023-12-31
Adding stereo_msgs from common_interfaces package

samehmohamed88:
This Pull Request adds the following two messages from the common_interfaces package:

1. shape_msgs
2. stereo_msgs

### DIFF
--- a/repositories/common_interfaces.BUILD.bazel
+++ b/repositories/common_interfaces.BUILD.bazel
@@ -161,6 +161,97 @@ py_library(
 )
 
 ros2_interface_library(
+    name = "shape_msgs",
+    srcs = glob([
+        "shape_msgs/msg/*.msg",
+    ]),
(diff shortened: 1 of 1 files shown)


repositories/common_interfaces.BUILD.bazel
```diff
@@ -160,6 +160,97 @@ py_library(
...
+    ]),
+    visibility = ["//visibility:public"],
+    deps = [
+        ":geometry_msgs",
+    ],
+)
+
+cpp_ros2_interface_library(
+    name = "_cpp_shape_msgs",
+    visibility = ["//visibility:private"],
```
mvukov, line 176: Here and down below: the default visibility is private, so, this qualifier can be removed for all targets with names beginning with an underscore.

mvukov:
There is some weirdness with permissions on the CI, I need to look at that. Otherwise the PR is solid.

PR #238, opened 2024-01-08
Bzlmod

lalten:
Add support for Bzlmod, Bazel's current external dependency management system.


lalten:
Current state is that things are complaining about boost. I think rather than trying to use the WORKSPACE-only rules_boost it would be best to switch over to https://github.com/bazelbuild/bazel-central-registry/pull/1280#issuecomment-1902491968 (once that's ready).

mvukov:
Boost is used at a single place https://github.com/mvukov/rules_ros2/blob/68a2b811d40bbe1713325a2fdd1c99993ae1f7c9/repositories/resource_retriever.BUILD.bazel#L13 (IMO, that package is outdated but still needed by the foxglove bridge). The receiver class is fairly small and contained last thing I remember. So, I'd suggest to patch it and use std::shared_ptr; then we can remove the boost dep. I can try to do this over the weekend or you can try to do it if you have some spare time. I'd target this change against the main branch to keep this PR as small as possible.

lalten:
Here's the issues/todos as of now:

* asio should probably depend on openssl like it does in WORKSPACE, but it doesn't in Bzlmod.
* pybind_bazel is harder than necessary to integrate and causes issues about Python not being found in CI (works locally), might revert that and stick to the http_archive.
* I don't understand why but Python3.10 causes issues. 3.11 works mostly, but py_loader complains about a compilation mismatch then.
* Need to update bzlmod situation in examples. I have the feeling that users need to add a bunch of bazel_deps in their repo that are transitive dependencies of rules_ros2. That's kind of what bzlmod is supposed to avoid. Let's see how it works when we get there.

WORKSPACE
```diff
@@ -35,6 +37,10 @@ load(
 
 install_rules_ros2_pip_deps()
 
+load("@pybind11_bazel//:python_configure.bzl", "python_configure")
+
+python_configure(name = "local_config_python")
```
mvukov, line 42: Pass here the interpreter target python_interpreter_target and then the new config is going to use that custom interpreter i.s.o. trying to use the system one. This will fix the CI issues.

repositories/repositories.bzl
```diff
@@ -1,23 +1,30 @@
...
+    )
+
     maybe(
         http_archive,
         name = "rules_python",
-        sha256 = "d70cd72a7a4880f0000a6346253414825c19cdd40a28289bdf67b8e6480edff8",
-        strip_prefix = "rules_python-0.28.0",
-        url = "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.28.0.tar.gz",
+        sha256 = "d71d2c67e0bce986e1c5a7731b4693226867c45bfe0b7c5e0067228a536fc580",
+        strip_prefix = "rules_python-0.29.0",
```
mvukov, line 26: Upgrading the deps is not a point of this PR and makes review more difficult. So, please use versions as they are.
Please also skip reordering the deps in this PR, as it also makes review harder -- and that can be done in a follow-up as well.

repositories/repositories.bzl
```diff
@@ -29,74 +36,86 @@ def ros2_repositories():
...
     maybe(
         http_archive,
-        name = "com_google_googletest",
-        sha256 = "8ad598c73ad796e0d8280b082cebd82a630d73e73cd3c70057938a6501bba5d7",
-        strip_prefix = "googletest-1.14.0",
-        url = "https://github.com/google/googletest/archive/refs/tags/v1.14.0.tar.gz",
+        name = "fmt",
+        build_file = "@com_github_mvukov_rules_ros2//repositories:fmt.BUILD.bazel",
+        sha256 = "1250e4cc58bf06ee631567523f48848dc4596133e163f02615c97f78bab6c811",
+        strip_prefix = "fmt-10.2.1",
```
mvukov, line 84: FYI, last thing I know is that spdlog still uses an older 9.x fmt.

repositories/repositories.bzl
```diff
@@ -29,74 +36,86 @@ def ros2_repositories():
...
 
     maybe(
         http_archive,
         name = "tinyxml2",
         build_file = "@com_github_mvukov_rules_ros2//repositories:tinyxml2.BUILD.bazel",
-        sha256 = "3bdf15128ba16686e69bce256cc468e76c7b94ff2c7f391cc5ec09e40bff3839",
-        strip_prefix = "tinyxml2-10.0.0",
-        urls = ["https://github.com/leethomason/tinyxml2/archive/refs/tags/10.0.0.tar.gz"],
+        sha256 = "cc2f1417c308b1f6acc54f88eb70771a0bf65f76282ce5c40e54cfe52952702c",
+        strip_prefix = "tinyxml2-9.0.0",
```
mvukov, line 101: No really a reason to downgrade this, or is there? :)

mvukov (commented):
High level comments: let's leave version updates and reordering of deps for follow-up PRs to make the review easier.

Great job so far!

WORKSPACE
lalten: Ah, makes sense! btw always feel free to push to the branch :)

repositories/repositories.bzl
lalten: I'll extract the version changes into separate PRs at some point. The rules_python upgrade was required due to https://github.com/bazelbuild/rules_python/issues/1669.
The other deps aren't really re-ordered, I moved them to a different function. I do agree the diff is basically impossible to read though :P

repositories/repositories.bzl
lalten: That's what I thought, but
https://github.com/bazelbuild/bazel-central-registry/blob/90ea6ab1277d1f338db9204d56276409b812ac99/modules/spdlog/1.12.0/MODULE.bazel#L7 wants 10
and
https://github.com/bazelbuild/bazel-central-registry/blob/90ea6ab1277d1f338db9204d56276409b812ac99/modules/fmt/10.1.1/MODULE.bazel#L4 says the compatibility level is incompatible with 9,
so we can't mix the versions

repositories/repositories.bzl
lalten: The reason is to have the same version in bzlmod and workspace. tinyxml 10 isn't on BCR (yet?) https://registry.bazel.build/modules/tinyxml2

WORKSPACE
lalten: So for WORKSPACE passing the `python_interpreter_target` works great, but for MODULE.bazel https://github.com/pybind/pybind11_bazel/issues/66 is getting in the way. Looks like we need to stick with the solution that's on main now and can't get pybind via Bzlmod for now :(

lalten:
Getting closer...
Still getting this from the WORKSPACE version of `asio` but not in the MODULE.bazel one.
```
external/asio/include/asio/ssl/detail/openssl_types.hpp:23:10: fatal error: 'openssl/conf.h' file not found
#include <openssl/conf.h>
         ^~~~~~~~~~~~~~~~
1 error generated.
```
The reason is that Bzlmod asio is configured as header-only: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/asio/1.28.2/patches/add_build_file.patch

But in WORKSPACE we use the checked-in asio.BUILD that has https://github.com/mvukov/rules_ros2/blob/4b8441a96ccd5fba5cf613f9637de28f26928a13/repositories/asio.BUILD.bazel#L13-L15 to switch from header-only to lib compilation and (because of that) has to add https://github.com/mvukov/rules_ros2/blob/4b8441a96ccd5fba5cf613f9637de28f26928a13/repositories/asio.BUILD.bazel#L18

CI is green with this change, but that may just mean that there are no tests that really need ssl :)

Not sure what's the better approach here. But IMO WORKSPACE and MODULE.bazel should both do the same thing.

We still need OpenSSL for libcurl, which is not on BCR yet.

repositories/repositories.bzl
lalten: added in https://github.com/bazelbuild/bazel-central-registry/pull/1395

lalten: back at 10 now

repositories/repositories.bzl
lalten: https://github.com/gabime/spdlog/discussions/2995#discussioncomment-8303850

WORKSPACE
lalten: pybind11 is tamed now

lalten:
State of things:
* tests seem green
* curl and pybind11 are now bzlmoded properly, so there are no http_archive dependencies apart from the ros2 related repos
* rules_python is still being very difficult with the numpy header version alias thing
* need to do things in a loop over all python versions. this is really stupid but apparently common now ([debian_dependency_bazelizer](https://github.com/shabanzd/debian_dependency_bazelizer/blob/b9ac2226645b43279440d0b092e7ef200e88a9c3/MODULE.bazel#L24), [rules_appimage](https://github.com/lalten/rules_appimage/blob/main/MODULE.bazel#L39), [grpc](https://github.com/bazelbuild/bazel-central-registry/blob/f6798c5dff2dacadf719363d1679d24f039d5c77/modules/grpc/1.56.3.bcr.1/MODULE.bazel#L54))
* Minor but I needed to remove the python_version from the pre-commit as I don't have Python3.10 locally. Not specifying is OK, it is only really needed for node
* Workspace `asio`'s BUILD is switched to header-only mode to be the same as the Bzlmod version
* Some smaller version changes to be in sync with what bzlmod versions are available on the BCR

The numpy header thing is really annoying still and I haven't found a good way to work around this. Maybe we can just leave this for now.

I _think_ things should be working now. Would be happy about a review @mvukov.

mvukov (commented):
Thanks for all your efforts. I left a couple of questions/comments.

BUILD.bazel
```diff
@@ -12,3 +13,11 @@ exports_files([
...
     "requirements_lock.txt",
 ])
+
+DEFAULT_PYTHON_VERSION = "3.10"  # keep in sync with MODULE.bazel
+
+alias(
+    name = "rules_ros2_pip_deps_numpy_headers",
+    actual = ("@rules_ros2_pip_deps_%s_numpy//:headers" % DEFAULT_PYTHON_VERSION.replace(".", "")) if BZLMOD_ENABLED else "@rules_ros2_pip_deps_numpy//:headers",
+    visibility = ["//visibility:public"],
+)
```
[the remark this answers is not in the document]
lalten: yep... I'll ask on the Bazel Slack about this when I next get some time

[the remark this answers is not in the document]
lalten: https://github.com/bazelbuild/rules_python/pull/1903 may solve this

otiv-milan:
[1e16de8](https://github.com/mvukov/rules_ros2/pull/238/commits/1e16de8c3797d94b54ea589524fa7be0e45481d9) looks promising, thanks for the efforts. We shall wait for the new rules_python release, right?

lalten:
> [1e16de8](https://github.com/mvukov/rules_ros2/pull/238/commits/1e16de8c3797d94b54ea589524fa7be0e45481d9) looks promising, thanks for the efforts. We shall wait for the new rules_python release, right?

yup, it landed in https://github.com/bazelbuild/rules_python/commit/730a2e39bd2702910f28629d4583b3ec49f4ee5e, now we just need a release.

I'm still not happy with the iterating over Python versions in MODULE.bazel. That's basically https://github.com/bazelbuild/rules_python/issues/1708.

mvukov:
> I'm still not happy with the iterating over Python versions in MODULE.bazel. That's basically https://github.com/bazelbuild/rules_python/issues/1708.

The requirements lock file in this repo is resolved with Python 3.10. So, using that with other Python probably won't work as there are packages with pre-compiled extensions. Thoughts:
* rename the lock file to have suffix _3_10 + mention this in the docs such that people are aware of this.
* define python Bazel dep as dev dep, forcing users of rules_ros2 to define their own dependency on rules_python -- like we do it in no-blzmod case.

WDYT?

lalten:
> rename the lock file to have suffix _3_10 + mention this in the docs such that people are aware of this.

There's also a `multi_pip_parse`, although I've never used that.
It might be worth looking into alternative python dependency machinery, like rules_uv or rules_poetry or rules_pycross. I don't have much experience with those, but maybe they have some solutions for this.

> define python Bazel dep as dev dep, forcing users of rules_ros2 to define their own dependency on rules_python -- like we do it in no-blzmod case.

I don't really like the user experience of that

lalten:
> It might be worth looking into alternative python dependency machinery, like rules_uv or rules_poetry or rules_pycross.

It seems like https://github.com/oxidase/rules_poetry is indeed a nice solution! I have a POC locally that seems to work.

One small challenge was that it doesn't expose a whl (because there isn't _one_ anymore as the interpreter isn't resolved "yet") for numpy, so can't use the new whl_filegroup.
I don't know how I haven't thought of this earlier, but you can totally just have a http_archive of https://github.com/numpy/numpy/tree/main/numpy/_core/include/numpy and inject a BUILD exposing the headers as cc_library.

I'll try and get the WORKSPACE version onto rules_poetry as well, and then that can already be landed separately.
Once that's out the door I think there are no blockers left for the rules_ros2 bzlmod migration :)

lalten:
Oh -- there is no Workspace version of rules_poetry! @oxidase, any chance you'd like to still add that?

lalten:
@mvukov -- it would be a big change, but wdyt about just leaving Workspace behind and going full Bzlmod? Then this wouldn't be an issue.

lalten:
I take back the thing about a numpy http_archive being a good idea. Looks like some needed includes are generated (https://github.com/numpy/numpy/blob/v1.26.4/numpy/core/meson.build#L608-L614) and not part part of the sdist release. So you have to know which interpreter to target.

oxidase:
@lalten can do it during the next month. Anyways appreciate any PRs to the repo

ahans:
> @mvukov -- it would be a big change, but wdyt about just leaving Workspace behind and going full Bzlmod? Then this wouldn't be an issue.

Just weighing in as a `rules_ros2` user: The projects relying on `rules_ros2` that I (help) maintain are all still on `WORKSPACE`. Making them stay on an old version by dropping `WORKSPACE` support is nothing I would welcome.

As for the issues with Python: I haven't really looked into bzlmod yet, so not sure I understand the full picture. But moving away from `rules_python` would also be a big step and potentially lead to other issues that are hard to foresee. I made the switch from `rules_python` to `rules_pycross` in one project. While it looked great on first sight, there were a couple of surprising issues. I was able to work around those mostly, but I'm really not sure if for an upstream project as `rules_ros2` this is a good idea.

Making it explicit that we rely on Python 3.10 by mentioning this in the docs and renaming the lock file sounds like an acceptable approach to me. Folks can still patch things if they know what they're doing. And unless they know what they're doing, they shouldn't mess with the Python version anyway, since ROS 2 Humble officially only supports 3.10.

Anyway, awesome effort going on here. Thanks a lot for that!

ahans:
> since ROS 2 Humble officially only supports 3.10.

Looks like I have to take back that part. Ubuntu 20.04 is still supported as a Tier 3 platform, so Humble will be kept Python 3.8 compatible. But  doubt that has much relevance, so the part about having to know what one is doing remains valid. ;-)

oxidase:
> But moving away from rules_python would also be a big step and potentially lead to other issues that are hard to foresee.

Heads up about moving away from `rules_python`. From my understanding internal Python rules are deprecated and are moved to `rules_python` as Starlark-based definitions in Bazel 7. For example PyInfo is selected by [config.enable_pystar](https://github.com/bazelbuild/rules_python/blob/main/python/py_info.bzl#L21) with Starlark `PyInfo`  not compatible to builtin `PyInfo`. Moving away from  `rules_python` will lead to issues like [this](https://github.com/aspect-build/rules_py/issues/338) with cryptic messages
```
ERROR: poetry-demo/BUILD.bazel:3:10: in deps attribute of py_binary rule //:main:
'@@rules_poetry~~poetry~poetry//:absl-py' does not have mandatory providers:
'PyInfo' or 'PyVirtualInfo'. Since this rule was created by the macro 'py_binary',
the error might have been caused by the macro implementation
```

ahans:
> Heads up about moving away from `rules_python`. From my understanding internal Python rules are deprecated and are moved to `rules_python` as Starlark-based definitions in Bazel 7.

My wording was not precise enough: Instead of just "Moving away from `rules_python`", it should have been "Moving away from `rules_python` for Python dependency management". The project I mentioned that now uses `rules_pycross` also still uses `rules_python` for things such as `py_library`, but no `pip_parse` or generated `requirement` from `rules_python` is used. That part is handled by `rules_pycross`.

oxidase:
@lalten I've just pushed [0.3.6 release](https://github.com/oxidase/rules_poetry/releases/tag/v0.3.6).

Added two examples:
* [single repository with WORKSPACE](https://github.com/oxidase/rules_poetry/blob/main/examples/workspace/WORKSPACE)

* [multiple repositories and multi-versioned Python with WORKSPACE](https://github.com/oxidase/rules_poetry/blob/main/examples/workspace_rules_python/WORKSPACE)

mvukov:
> @mvukov -- it would be a big change, but wdyt about just leaving Workspace behind and going full Bzlmod? Then this wouldn't be an issue.

Hm, no. Including me, there are folks around depending on nobzlmod.

Regarding

> > define python Bazel dep as dev dep, forcing users of rules_ros2 to define their own dependency on rules_python -- like we do it in no-blzmod case.
> 
> I don't really like the user experience of that

In main branch, noblzmod still, I left rules_python setup open to users; intentionally. I don't really care which Python folks are going to use. As long as users provide necessary stuff, this ruleset is going to work: rules_ros2_pip_deps in particular. On the pip requirements side, there is a set of necessary deps defined in requirements.txt and this repo provides a lock file for Python 3.10. Users happy with that, fine. Not so much? They can resolve the deps for an interpreter version they want to work with.

Now, what can happen in the current state of this PR, is that folks can have rules_ros2_pip_deps and their own my_repo_pip_deps. Let's say that both pip_deps repos have a package "foo" resolved at different versions (foo-1.3 and foo-2.4 for instance) and used as a dep (e.g. transitively) in the same Bazel target. Best case you have the same package shipped two times, worst case you get creepy runtime errors due to some version incompatibility (this would happen sooner or later). In an extreme case, rules_ros2_pip_deps is resolved for one Python version and my_repo_pip_deps for another. To avoid those issues, I strongly suggest this repo defines rules_python and rules_ros2_pip_deps as dev deps, forcing users defining rules_ros2_pip_deps in their own repo. This isn't documented yet (should be :) ), but folks should ideally have rules_ros2_pip_deps where they have both necessary packages for rules_ros2 and other packages they might need.

Those are just my thought, pretty much how I'd do it. I am very open to hear other suggestions and learn how users of rules_ros2 set up Python in their own monorepos.

lalten:
I just learned about https://rules-python.readthedocs.io/en/latest/toolchains.html#pinning-to-a-python-version, that is likely what we want here.

(still need to update this PR with the rules_python release)

lalten:
Things seem to work, the only drawback is that `load("@rules_python//python:defs.bzl", "py_binary")` must become `load("@rules_ros2_pythons//3.10:defs.bzl", "py_binary")`, so there is a bunch of `3.10` hardcoded everywhere. While not pretty I think that would be acceptable

lalten:
I am unsure what causes the test failures (or how to fix them), could you have a look @mvukov or @ahans ?

mering:
> Things seem to work, the only drawback is that `load("@rules_python//python:defs.bzl", "py_binary")` must become `load("@rules_ros2_pythons//3.10:defs.bzl", "py_binary")`, so there is a bunch of `3.10` hardcoded everywhere. While not pretty I think that would be acceptable

Why do you think this is necessary? AFAIK `rules_python` will use the default toolchain when using `load("@rules_python//python:defs.bzl", "py_binary")`, so `load("@rules_ros2_pythons//3.10:defs.bzl", "py_binary")` should be equivalent to registering the `3.10` toolchain as default.

MODULE.bazel
```diff
@@ -0,0 +1,102 @@
...
+bazel_dep(name = "tinyxml2", version = "10.0.0")
+bazel_dep(name = "websocketpp", version = "0.8.2")
+bazel_dep(name = "yaml-cpp", version = "0.8.0")
+bazel_dep(name = "zlib", version = "1.3")
+bazel_dep(name = "zstd", version = "1.5.5.bcr.1")
+
+_DEFAULT_PYTHON = "3.10"
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(python_version = _DEFAULT_PYTHON)
```
mering, line 32: Add `is_default=True`.

lalten:
> > Things seem to work, the only drawback is that `load("@rules_python//python:defs.bzl", "py_binary")` must become `load("@rules_ros2_pythons//3.10:defs.bzl", "py_binary")`, so there is a bunch of `3.10` hardcoded everywhere. While not pretty I think that would be acceptable
> 
> Why do you think this is necessary? AFAIK `rules_python` will use the default toolchain when using `load("@rules_python//python:defs.bzl", "py_binary")`, so `load("@rules_ros2_pythons//3.10:defs.bzl", "py_binary")` should be equivalent to registering the `3.10` toolchain as default.

right, but my understanding is that if the root repo uses a default != 3.10, the rules_ros2_pip_deps are not guaranteed to be compatible

mering:
> > > Things seem to work, the only drawback is that `load("@rules_python//python:defs.bzl", "py_binary")` must become `load("@rules_ros2_pythons//3.10:defs.bzl", "py_binary")`, so there is a bunch of `3.10` hardcoded everywhere. While not pretty I think that would be acceptable
> > 
> > 
> > Why do you think this is necessary? AFAIK `rules_python` will use the default toolchain when using `load("@rules_python//python:defs.bzl", "py_binary")`, so `load("@rules_ros2_pythons//3.10:defs.bzl", "py_binary")` should be equivalent to registering the `3.10` toolchain as default.
> 
> right, but my understanding is that if the root repo uses a default != 3.10, the rules_ros2_pip_deps are not guaranteed to be compatible

Are you sure that they will chosen in the first place? My understanding is that it will only use pip deps with the correct version. This is why most projects specify the same pip deps for many Python versions. IIUC when it won't find pip deps in the correct version, it will throw an error.

udaya2899:
@lalten were you able to make progress on this PR? :)

udaya2899:
No worries, thank you for your contribution. As far as I understand, you got everything working already, right? Maybe time to merge 👀 ?

lalten:
Tried to update this branch (still waiting on https://github.com/mvukov/rules_ros2/pull/344 and  https://github.com/mvukov/rules_ros2/pull/340) but the new rust features require some work to be compatible with Bzlmod.

In particular it's not obvious how to get the bindgen stuff into Bzlmod

ahans:
@lalten @mvukov Anything I can lend a hand with? Being able to move to bzlmod would be big!

lalten:
My previous comment regarding the rust stuff ist still the state of things. I'm thinking if we could add a custom build config flag or so to disable the rust support that might be a way around those issues.

ahans:
> My previous comment regarding the rust stuff is still the state of things. I'm thinking if we could add a custom build config flag or so to disable the rust support that might be a way around those issues.

Yeah, that sounds good. You could tag all Rust-related tests with `no_bzlmod` and then use `--tag_test_filters` to exclude them. Or simply exclude them in the CI config by doing something like `bazel test -- //... -//ros2/test/rust/...`. We'd need some documentation making clear that Rust support is only available in the WORKSPACE version, but that should be about it? The bindgen discussion sounds like it could be a while until this is fully supported with bzlmod... @mvukov What do you think?

mvukov:
I have no objections merging bzlmod support. But, workspace-based workflow must work, no compromises on that one for the time being. I am not personally using bzlmod (yet) in any projects, so, no Rust support is fine.

Something to think about: https://github.com/mvukov/rules_ros is for instance migrated to bzlmod recently. Since no one is using the workspace workflow, it was decided to remove workspace support. Something I found along the way:
* Python toolchain to be used is the one from the parent project, i.e. rules_ros.
* In this repo we have rules_ros2_pip_deps, in the other rules_ros_pip_deps. Turns out one can not  override this repo from his/her own repo. The issue is tracked in https://github.com/bazelbuild/rules_python/issues/1791. Depending on what parts of and how you use rules_ros2, this might be a (big) problem. In workspace based workflow one can extend requirements.txt with their own deps and use that both with rules_ros2 in their own repo. (Of course, one needs to be careful not to break things.) This is a blocker IMO to remove workspace workflow from rules_ros2.

mvukov:
https://github.com/bazelbuild/rules_python/issues/2370 also very relevant and would solve issues with bzlmod and python packages coming from multiple places.

mvukov:
rust bindgen is available in bzlmod iirc since rules_rust 0.55.0.

kgreenek:
Watching this eagerly! I'm going through the painful process now of getting my company's dependencies all up on BCR / working with bazel modules. Bazel 8 disabling the workspace option by default put a fire under us as motivation.

kgreenek:
I have been giving this branch a try in my repo and so far it's working!

I did run into a potential issue though, that looks like it might just be a typo.

It looks like there is a reference somewhere to `@tinyxml`, which might just need to be changed to `@tinyxml2`?

Here is the error I get when I try to build the urdf rule:

```
ERROR: /home/kgk/.cache/bazel/_bazel_kgk/ec654fdc5ba6a08f3050c551a6e071c3/external/com_github_mvukov_rules_ros2~~non_module_deps~ros2_urdfdom/BUILD.bazel:6:11: no such package '@@[unknown repo 'tinyxml' requested from @@com_github_mvukov_rules_ros2~~non_module_deps~ros2_urdfdom (did you mean 'tinyxml2'?)]//': The repository '@@[unknown repo 'tinyxml' requested from @@com_github_mvukov_rules_ros2~~non_module_deps~ros2_urdfdom (did you mean 'tinyxml2'?)]' could not be resolved: No repository visible as '@tinyxml' from repository '@@com_github_mvukov_rules_ros2~~non_module_deps~ros2_urdfdom' and referenced by '@@com_github_mvukov_rules_ros2~~non_module_deps~ros2_urdfdom//:urdf_parser'
```

I'll also keep you updated if I run into other issues. Thanks for putting this PR up.

kgreenek:
I sent up a PR to add tinyxml to the BCR: https://github.com/bazelbuild/bazel-central-registry/pull/3520

MODULE.bazel
```diff
@@ -0,0 +1,137 @@
...
+bazel_dep(name = "pybind11_bazel", version = "2.12.0")
+bazel_dep(name = "re2", version = "2024-07-02")  # this is only here to force graph to resolve to higher res2 version compatible with pybind11_bazel above. Remove once no longer needed.
+bazel_dep(name = "readerwriterqueue", version = "1.0.6")
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "rules_foreign_cc", version = "0.11.1")
+bazel_dep(name = "rules_python", version = "0.35.0")
+bazel_dep(name = "rules_rust", version = "0.49.3")
+bazel_dep(name = "spdlog", version = "1.14.1")
+bazel_dep(name = "sqlite3", version = "3.42.0.bcr.1")
+bazel_dep(name = "tinyxml2", version = "10.0.0")
```
kgreenek, line 22:
```suggestion
bazel_dep(name = "tinyxml", version = "2.6.2")
bazel_dep(name = "tinyxml2", version = "10.0.0")
```

kgreenek: I updated my repo and verified this fix worked!

lalten:
Updated the PR, there's still issues with rust: https://github.com/mvukov/rules_ros2/actions/runs/12713955456/job/35443024338?pr=238#step:5:48 Not sure what the issue is here.

Also the examples fail with an equally cryptic error about cycles in Workspace (even though it should use Bzlmod?): https://github.com/mvukov/rules_ros2/actions/runs/12713955456/job/35443025705?pr=238#step:5:32

lalten:
It seems things are working now. @mvukov please have another look. Would be great to lift rules_ros2 into the Bzlmod age.

.github/workflows/main.yml
```diff
@@ -47,24 +47,33 @@ jobs:
     name: Run tests
     needs: configure
     runs-on: ubuntu-22.04
+    timeout-minutes: 600
```
lalten, line 50: This is the [usage limit](https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits) for job execution times, but it [looks like](https://github.com/mvukov/rules_ros2/actions/runs/12966673139/job/36167701033?pr=238) it's not enough to get the llvm dependency built in time with clang :/

udaya2899: Another alternative and a better long term solution is to use a Remote Build Execution service like buildbuddy.io (they have a free tier for open source projects) and the CI builds could become much much faster (hopefully in a matter of minutes, not hours).

I can help set it up some time if that's really slowing us down here.

mvukov:
Yes, I know that clang provided from rules_rust takes a lot of time. For this purpose I route system clang https://github.com/mvukov/rules_ros2/blob/main/repositories/rust_setup_stage_2.bzl#L10-L15, which is fine for a number of use-cases.

mvukov:
We should do something similar with bzlmod approach.

lalten:
> Yes, I know that clang provided from rules_rust takes a lot of time. For this purpose I route system clang https://github.com/mvukov/rules_ros2/blob/main/repositories/rust_setup_stage_2.bzl#L10-L15, which is fine for a number of use-cases.

We don't have any system clang on our side so that would require patching which I wouldn't love.
Wdyt about enabling caching not just on main but also on PR branches? Then it would pass on retry.

ahans:
> > Yes, I know that clang provided from rules_rust takes a lot of time. For this purpose I route system clang https://github.com/mvukov/rules_ros2/blob/main/repositories/rust_setup_stage_2.bzl#L10-L15, which is fine for a number of use-cases.
> 
> We don't have any system clang on our side so that would require patching which I wouldn't love. Wdyt about enabling caching not just on main but also on PR branches? Then it would pass on retry.

Can't we use a system-provided clang optionally? If something is found, we set up the toolchain. Otherwise, we leave the default and build it from source.

Also, when exactly is this being built? Only when Rust-related rules are being used? If not, I would also have to patch this for our usecase. We don't need any Rust support, so building clang locally would be a huge waste and something I can't sell my users. Without remote caching, everyone would have to build this at least once.

lalten:
I looked into making rust/llvm optional. It doesn't seem easy but could be possible if upstream rules_rust_bindgen adds support for using a host clang.

* Generally it is possible to provide build flags that switch out dependencies or configure build time constants with select()s (https://github.com/bazelbuild/bazel-central-registry/blob/2119aab860f062eb361aa63d819c19b00304f72e/modules/curl/8.8.0.bcr.2/patches/add_build_file.patch#L15-L48). However
* to provide support for rust_bindgen in [repositories/ros2_rust.BUILD.bazel](https://github.com/mvukov/rules_ros2/blob/64a43884728228f38dde76e41ce9199697604490/repositories/ros2_rust.BUILD.bazel#L35) we need to be able to `load("@rules_rust_bindgen//` at analysis time already,
* this means `@rules_rust_bindgen` must be part of MODULE.bazel (if the ros2_rust.BUILD is not patched away)
* The rules_rust_bindgen bazel_dep [requires the `llvm-project` bazel_dep](https://github.com/bazelbuild/bazel-central-registry/blob/2119aab860f062eb361aa63d819c19b00304f72e/modules/rules_rust_bindgen/0.57.1/MODULE.bazel#L28-L31) but then 
* rules_rust_bindgen only actually uses the `@llvm-project//clang` and `@llvm-project//clang:libclang` targets directly, in its [BUILD.bazel](https://github.com/bazelbuild/rules_rust/blob/d256c7bb4668bbdfd1e25971531c874c23f4646a/extensions/bindgen/BUILD.bazel#L29-L37)
* That means if rules_rust_bindgen used a config_setting to select() its clang target dependency which would be configurable to point at a non-hermetic host clang this might work.

lalten:
I'm not sure why rules_rust_bindgen even depends on `llvm-project` when a pre-compiled `llvm_toolchain` would be enough? With this patch in rules_rust `cd extensions/bindgen; bazel test //...` works there without compiling all of llvm.
I guess I'll suggest this oder a label_flag based solution upstream when I find time.

```diff
diff --git a/extensions/bindgen/BUILD.bazel b/extensions/bindgen/BUILD.bazel
index aecae999..3a9f9e5c 100644
--- a/extensions/bindgen/BUILD.bazel
+++ b/extensions/bindgen/BUILD.bazel
@@ -28,12 +28,25 @@ bzl_library(
 
 cc_stdcc17_transitioned_binary(
     name = "clang",
-    target = "@llvm-project//clang:clang",
+    target = ":clang_bin",
+)
+
+genrule(
+    name = "clang_bin",
+    srcs = ["@llvm_toolchain_llvm//:bin/clang"],
+    outs = ["clang_bin"],
+    cmd = "ln -s $(location @llvm_toolchain_llvm//:bin/clang) $@",
+    executable = True,
 )
 
 cc_stdcc17_transitioned_library(
     name = "libclang",
-    target = "@llvm-project//clang:libclang",
+    target = ":clang_lib",
+)
+
+cc_library(
+    name = "clang_lib",
+    srcs = ["@llvm_toolchain_llvm//:libclang"],
 )
 
 rust_bindgen_toolchain(
diff --git a/extensions/bindgen/MODULE.bazel b/extensions/bindgen/MODULE.bazel
index a84f7d06..125dbafc 100644
--- a/extensions/bindgen/MODULE.bazel
+++ b/extensions/bindgen/MODULE.bazel
@@ -29,6 +29,15 @@ bazel_dep(
     name = "llvm-project",
     version = "17.0.3.bcr.1",
 )
+bazel_dep(name = "toolchains_llvm", version = "1.3.0")
+
+llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
+llvm.toolchain(llvm_version = "17.0.6")
+
+use_repo(llvm, "llvm_toolchain")
+use_repo(llvm, "llvm_toolchain_llvm")
+
+register_toolchains("@llvm_toolchain//:all")
 
 rust_ext = use_extension("//:extensions.bzl", "rust_ext")
 use_repo(
```

lalten:
@mvukov I disabled the rust tests in the Bzlmod CI for now, which means there is no llvm-project dependency anymore and the jobs actually finish green. Depending on https://github.com/bazelbuild/rules_rust/pull/3231 the tests could just use a pre-compiled clang instead in the future.

IMHO the current state could be landed already. There will always be more things to improve but it's so much easier to make small iterations off main instead of a long-living branch like this.

ahans:
This is great, awesome work @lalten! I just integrated this into my bzlmod test repo and everything worked out of the box! I basically followed what you do in `examples/MODULE.bazel` on your branch. I was wondering if it's possible to move the long list of repos in `use_repo` to some `.bzl` file or something like that? So that `MODULE.bazel` is less cluttered. Sorry if that's a stupid question, I'm still very new to bzlmod ...

lalten:
> I was wondering if it's possible to move the long list of repos in `use_repo` to some `.bzl` file or something like that? So that `MODULE.bazel` is less cluttered.

You can use https://bazel.build/rules/lib/globals/module#include with Bazel 7.2+

MODULE.bazel
```diff
@@ -0,0 +1,153 @@
...
+)
+use_repo(pip, "rules_ros2_pip_deps")
+
+dev_pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip", dev_dependency = True)
+dev_pip.parse(
+    hub_name = "rules_ros2_resolver_deps",
+    python_version = _DEFAULT_PYTHON,
+    requirements_lock = "//repositories/private:resolver_requirements_lock.txt",
+)
+use_repo(dev_pip, "rules_ros2_resolver_deps")
```
mering, line 71: It would be great if this could support multiple Python versions. A common approach is to define a list of supported Python versions and use list comprehension for `python.toolchain()` and `pip.parse()`: https://github.com/bazelbuild/bazel-central-registry/blob/40d0287868ab53ac511a8af993000f24f6b6e1fe/modules/grpc/1.70.1/MODULE.bazel#L50-L80

lalten: I don't love that approach (I used it in rules_appimage for a while) as it will cause warnings on every single build.
For this resolver pypi it doesn't matter btw as this is an internal tool and not exposed by the module.
For the other py_binaries in this repo it's not simple to do either because of the bumpy wheel dep which must pin a specific Python header. It can't be compatible with multiple Pythons. In fact with the current change set of this PR the py_binaries will bring their own interpreter because of that.

mering: rules_python allows using current headers target. So whatever default version the root module picks, it should just work.

mering: Or with numpy, it should be good enough to specify a version and Bazel should just download the correct wheel from pypi.org.

lalten: Unfortunately not in this case, the numpy headers are python version specific and are extracted from the wheel. They are generated at numpy build time.

lalten: https://github.com/mvukov/rules_ros2/blob/main/ros2%2FBUILD.bazel#L94-L98

mering: I don't see a problem as long as the pip wheel and the python headers use the same version of Python (the default version). I think it should just work.

ahans:
We use both launch as well as launch testing, but haven't migrated to bzlmod yet. I'm currently in the process of doing that and will report back my findings. Though not the preferred solution, I'm fine with maintaining local patches as well. If that is necessary to work around Python issues, so be it.

kgreenek:
I have been testing this for a few weeks now. ros2 launch seems to be working for me just fine!

The only issue I'm running into is with foxglove_bridge. It might be something with my setup, but I'm bringing it up here in case anybody else is seeing this too.

foxglove_bridge runs just fine following the example in the documentation. However, when I try to open Foxglove Studio and connect, foxglove shows that the connection failed. Looking at the logs for foxglove_bridge I see errors like the following:

```
[foxglove_bridge_node-1] [INFO] [1738873667.181369350] [foxglove_bridge]: [WS] asio handle_timer error: asio.system:125 (Operation aborted.)
[foxglove_bridge_node-1] [ERROR] [1738873667.181450158] [foxglove_bridge]: [WS] handle_transport_init received error: invalid state
[foxglove_bridge_node-1] [ERROR] [1738873667.181508330] [foxglove_bridge]: [WS] handle_read_handshake error: asio.system:11 (Resource temporarily unavailable)
```

ros2/plugin_aspects.bzl
```diff
@@ -38,15 +38,16 @@ Ros2PluginCollectorAspectInfo = provider(
     ],
 )
 
-_ROS2_COLLECTOR_ATTR_ASPECTS = ["data", "deps"]
+# From https://github.com/bazelbuild/rules_python/blob/0.33.2/python/config_settings/transition.bzl#L121-L140
```
mvukov, line 41: I just updated rules_python to 1.1.0. main branch should be merged in here.

ros2/test/rust/BUILD.bazel
```diff
@@ -49,6 +49,7 @@ rust_binary(
 ros2_test(
     name = "tests",
     size = "small",
+    flaky = True,
```
mvukov, line 52: Not quite related but can go, I know it's flaky.

third_party/foxglove_bridge/BUILD.bazel
```diff
@@ -42,6 +42,7 @@ ros2_cpp_test(
     size = "small",
     srcs = ["foxglove_bridge_tests.cc"],
     env = {"ROS_DISTRO": "humble"},
+    flaky = True,
```
mvukov, line 45: Fine here as well.

mvukov (commented):
Why do you replace `load("@rules_python//python:defs.bzl", "py_library")` with `load("@rules_ros2_pythons//3.10:defs.bzl", "py_binary")`? Is that abs needed in this PR? If not abs needed, please revert those changes. We can debate this in a follow-up.

Do we need WORKSPACE.bzlmod?

Let's keep the amount of changes minimal and easy to review.

Great work!

mvukov:
> I have had to maintain a couple of local patches for other external dependencies that use pip with rules_python. I patch their requirements lock files to match the versions required by my repo, which is kept compatible with rules_ros2. It's a bit painful, but I don't know of a better way until rules_python improves the situation.

It's gross. But there is no way around it until https://github.com/bazelbuild/rules_python/issues/1791 is sorted out.

mvukov:
>  Wdyt about enabling caching not just on main but also on PR branches? Then it would pass on retry.

For remote caching a token is needed. Github prevents using repo secrets in PRs. If you want the write access to this git remote, lemme know.

mering:
> It's gross. But there is no way around it until [bazelbuild/rules_python#1791](https://github.com/bazelbuild/rules_python/issues/1791) is sorted out.

IIUC This will never be fixed as it is based on the foundations of how rules_python works: It's constructing a PYTHONPATH env var for all dependencies. When the same dependency (package name) appears twice in the dependency graph, the order decides which one is taken.

.github/workflows/main.yml
```diff
@@ -89,13 +98,19 @@ jobs:
...
           XDG_CACHE_HOME: /github/home/.cache/bazel-repo
           BUILDBUDDY_ORG_API_KEY: ${{ secrets.BUILDBUDDY_ORG_API_KEY }}
           USER: ${{ needs.configure.outputs.user_name }}
         working-directory: ${{ matrix.folder }}
-        run: ${GITHUB_WORKSPACE}/.github/workflows/test.sh ${{ matrix.config_option }}
+        run: >
+          ${GITHUB_WORKSPACE}/.github/workflows/test.sh
+          ${{ matrix.config_option }}
+          ${{ matrix.bzlmod_flags }}
+          ${{ matrix.external_dependency_system == 'bzlmod' && matrix.folder == '.' && '-- -//ros2/test/rust/...' || '--'}}
```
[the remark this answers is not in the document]
lalten: not possible until https://github.com/bazelbuild/rules_rust/pull/3231

.bazelrc
```diff
@@ -4,13 +4,9 @@
...
-# This is mandatory if your build env doesn't have a proper Python 3 exe
-# -- like the CI build image for this repo.
-# The build image is using a fake python exe, see
-# https://github.com/mvukov/bazel_builder/blob/main/Dockerfile.
-# If blzmod is enabled, but not used as in this repo at the moment),
-# then bazel for some reason tries to auto-detect a Python exe.
-build --noenable_bzlmod
+# Don't let local Python site packages leak into the build and cause problems
+common --incompatible_strict_action_env
+common --action_env=PYTHONNOUSERSITE=1
```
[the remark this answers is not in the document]
lalten: removed. must have been some merge hickup

.github/workflows/main.yml
```diff
@@ -47,24 +47,33 @@ jobs:
...
     needs: configure
     runs-on: ubuntu-22.04
+    timeout-minutes: 600
     container:
       image: mvukov/bazel-builder:commit-32e5440bfdfa795c7fdef8885374e5cbbb9976b5
       options: --user ${{ needs.configure.outputs.uid_gid }}
 
     # Run bazel test with gcc and clang in each workspace
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
lalten: flipped

.gitignore
```diff
@@ -2,4 +2,5 @@
 bazel-*
 __pycache__
 user.bazelrc
+external/
```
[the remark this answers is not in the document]
lalten: this was for a symlink to the bazel external folder which is super helpful for debugging. removed.

external
```diff
@@ -0,0 +1 @@
+bazel-out/../../../external
```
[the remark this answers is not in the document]
lalten: removed

repositories/foxglove_bridge.BUILD.bazel
```diff
@@ -25,6 +25,7 @@ cc_library(
     copts = ["-std=c++17"],
     includes = ["foxglove_bridge_base/include"],
     deps = [
+        "@boringssl//:ssl",
```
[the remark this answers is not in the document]
lalten: can split off to a separate one. I don't have full memory of how this failed but a quick look at the srcs shows that it indeed depends on (open)ssl symbols: https://github.com/foxglove/ros-foxglove-bridge/blob/e4af8e156b99bdb51135a0612d97436cc355049f/foxglove_bridge_base/src/server_factory.cpp#L51 so it is correct to specify the dependency here

repositories/repositories.bzl
```diff
@@ -32,9 +32,9 @@ def ros2_workspace_repositories():
         http_archive,
         name = "fmt",
         build_file = "@com_github_mvukov_rules_ros2//repositories:fmt.BUILD.bazel",
-        sha256 = "5dea48d1fcddc3ec571ce2058e13910a0d4a6bab4cc09a809d8b1dd1c88ae6f2",
-        strip_prefix = "fmt-9.1.0",
-        url = "https://github.com/fmtlib/fmt/archive/9.1.0.tar.gz",
+        sha256 = "1250e4cc58bf06ee631567523f48848dc4596133e163f02615c97f78bab6c811",
+        strip_prefix = "fmt-10.2.1",
```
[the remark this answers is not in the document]
lalten: this is to get Workspace and Bzlmod to the same version. I don't have full memory but there was some issue with https://registry.bazel.build/modules/fmt lower than this. In any case it definitely works, that's what we use internally for a while.

repositories/repositories.bzl
```diff
@@ -155,12 +155,9 @@ def ros2_workspace_repositories():
...
         name = "zlib",
         build_file = "@com_github_mvukov_rules_ros2//repositories:zlib.BUILD.bazel",
-        sha256 = "d14c38e313afc35a9a8760dadf26042f51ea0f5d154b0630a31da0540107fb98",
-        strip_prefix = "zlib-1.2.13",
-        urls = [
-            "https://github.com/madler/zlib/releases/download/v1.2.13/zlib-1.2.13.tar.xz",
-            "https://zlib.net/zlib-1.2.13.tar.xz",
-        ],
+        sha256 = "8a9ba2898e1d0d774eca6ba5b4627a11e5588ba85c8851336eb38de4683050a7",
+        strip_prefix = "zlib-1.3",
```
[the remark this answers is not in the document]
lalten: same, to get BCR and Workspace zlib onto the same version. I think 1.2.13 was yanked for security issues: https://registry.bazel.build/modules/zlib

ros2/plugin_aspects.bzl
lalten: done

.pre-commit-config.yaml
```diff
@@ -1,5 +1,4 @@
 default_language_version:
-  python: python3.10
```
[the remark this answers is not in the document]
lalten: Ah, this didn't work on a laptop that with Arch Linux installed: specifying 3.10 here meant I would have to install Python from source :P I'll add it back in.

MODULE.bazel
```diff
@@ -0,0 +1,153 @@
...
+)
+crate.from_specs(
+    name = "rules_ros2_crate_index",
+    # Add these attrs when crate.from_specs supports them (rules_rust 0.57.0)
+    # cargo_lockfile = "//repositories/rust:Cargo.lock",
+    # lockfile = "//repositories/rust:Cargo.Bazel.lock",
+)
+use_repo(crate, "rules_ros2_crate_index")
+
+non_module_deps = use_extension("//:extensions.bzl", "non_module_deps")
```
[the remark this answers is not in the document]
lalten: Nope, users have to explicitly list the repos they want visible. Usually the list would be a lot smaller though. It's just that this repo itself uses _all_ the things.

examples/MODULE.bazel
```diff
@@ -0,0 +1,58 @@
...
+    module_name = "com_github_mvukov_rules_ros2",
+    path = "..",
+)
+
+bazel_dep(name = "rules_cc", version = "0.0.17")
+bazel_dep(name = "rules_python", version = "0.35.0")
+bazel_dep(name = "rules_rust", version = "0.56.0")
+
+non_module_deps = use_extension("@com_github_mvukov_rules_ros2//:extensions.bzl", "non_module_deps")
+use_repo(
```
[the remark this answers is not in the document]
lalten: list is small now, not needed I think

extensions.bzl
```diff
@@ -0,0 +1,8 @@
+load("@com_github_mvukov_rules_ros2//repositories:repositories.bzl", "ros2_repositories")
```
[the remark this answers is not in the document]
lalten: done

BUILD.bazel
```diff
@@ -1,4 +1,4 @@
-load("@rules_python//python:pip.bzl", "compile_pip_requirements")
+load("@rules_ros2_pythons//3.10:defs.bzl", "compile_pip_requirements")
```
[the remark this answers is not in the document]
lalten: I checked, and no -- because this isn't actually run at build time. removed

lalten:
I removed the "multi" version handling for Python now and added notes in the example WORKSPACE and MODULE.bazel about the requirement to pin the Python version to 3.10 to avoid numpy import problems in WORKSPACE. In Bzlmod, Bazel detects the misconfiguration and fails in analysis phase already

lalten:
Supporting something else than 3.10 would be neat but I feel it's out of scope for this PR, especially given that Workspace doesn't support any other version either.

WORKSPACE
```diff
@@ -10,17 +10,18 @@ load("//repositories:deps.bzl", "ros2_deps")
...
 
 py_repositories()
 
-python_register_toolchains(
-    name = "rules_ros2_python",
-    python_version = "3.10",
+python_register_multi_toolchains(
+    name = "rules_ros2_pythons",
+    default_version = "3.10",
+    python_versions = ["3.10"],
```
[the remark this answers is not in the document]
lalten: removed

examples/WORKSPACE
```diff
@@ -26,17 +26,18 @@ load("@com_github_mvukov_rules_ros2//repositories:deps.bzl", "ros2_deps")
 
 ros2_deps()
 
-load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
+load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_multi_toolchains")
```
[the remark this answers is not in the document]
lalten: removed

ahans:
Update from my side: Things are mostly working. The only patching I had to do was for `pip.parse`, where I give `experimental_index_url` and `download_only`, so that cross compilation works.

However, being able to customize this more, similar to what we had in WORKSPACE, would be nice. I'm currently trying to also migrate to rules_py, and that is less straightforward than I had hoped. There may be issues with interop of rules from rules_py with standard rules_python. But I think what I'm seeing is unrelated to this: When I have a target that uses something from rules_ros2 that depends on numpy and also uses numpy itself via the main repo's pypi, I end up with two copies. They're the same version, so maybe I can just silence the warning and forget about it. But maybe not. So the question is: If I wanted to provide Python deps from the main repo to rules_ros2, how could I do that? Or is there some other solution to this?

mering:
> If I wanted to provide Python deps from the main repo to rules_ros2, how could I do that? Or is there some other solution to this?

You can use the Python deps from `rules_ros2` by using the `use_repo()` call on the module extension and use this repo as `deps`:
```starlark
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
use_repo(pip, "rules_ros2_pip_deps")
```

mering:
> Any chance we can do this the other way around, i.e., make some `pypi` from the main module available in `rules_ros2`?

Unfortunately, no as this would require `rules_ros2` to add a dependency to your module into their MODULE.bazel.

rules_python does not support mixing multiple requirements.txt for a single target (they call it dependency closure). Some more details can be found in https://github.com/bazelbuild/rules_python/issues/2370.

ahans:
> > Any chance we can do this the other way around, i.e., make some `pypi` from the main module available in `rules_ros2`?

> Unfortunately, no as this would require `rules_ros2` to add a dependency to your module into their MODULE.bazel.

What if I patch that in? I have to do some patching anyways to support cross building.

examples/MODULE.bazel
```diff
@@ -0,0 +1,36 @@
...
+
+bazel_dep(name = "com_github_mvukov_rules_ros2")
+local_path_override(
+    module_name = "com_github_mvukov_rules_ros2",
+    path = "..",
+)
+
+bazel_dep(name = "rules_cc", version = "0.0.17")
+bazel_dep(name = "rules_python", version = "1.1.0")
+bazel_dep(name = "rules_rust", version = "0.56.0")
```
mvukov, line 11: Why do we need this when rules_rust is not going to be available from bzlmod for the time being?

.github/workflows/main.yml
mvukov: This is not needed any more IMO as ros2_rust is not available with bzlmod.

examples/WORKSPACE
```diff
@@ -32,6 +32,10 @@ py_repositories()
 
 python_register_toolchains(
     name = "rules_ros2_python",
+    # Note that it is required to pin this specific version because of dependencies to numpy headers from rules_ros2's 3.10 numpy dep.
+    # Otherwise you will get errors at runtime (not build time!) like:
+    #     ModuleNotFoundError: No module named 'numpy.core._multiarray_umath'
+    #     Importing the numpy C-extensions failed.
```
mvukov, line 38: Not quite accurate. We need first because the provided `requirements_lock.txt ` is resolved for Python 3.10. Then we need that because of numpy. It should be formulated that one needs to keep the python_version and the resolved lock file in sync using the same Python version.

.bazelrc
```diff
@@ -4,14 +4,6 @@
 # Mandatory at the moment for Bazel 7.0.0.
 build --noincompatible_sandbox_hermetic_tmp
 
-# This is mandatory if your build env doesn't have a proper Python 3 exe
```
mvukov, line 7: workspace-based workflow will remain as default, as that's the one fully functional.

I realized we need good documentation now to navigate new users and less experienced with Bazel: workspace vs bzlmod. Default workspace and clearly state that all advertised features are functional. Then bzlmod with decent notes on caveats.

examples/MODULE.bazel
lalten: Rust support is fully functional in Bzlmod and tests are green if you manage to run them. The only issue is that the build takes longer than the max GitHub Actions timeout. Locally there is no issue and all tests pass.
Build time can be reduced by using the same host clang as in Workspace. I have a local git stash to do that, but it can only work after https://github.com/bazelbuild/rules_rust/pull/3231 lands and is published to BCR. (because git overrides in rules_ros2 don't have effects in consumers of rules_ros2)

.bazelrc
lalten: Both variants are fully functional and these .bazelrc lines are not needed in either anymore

examples/WORKSPACE
lalten: I guess you could say that yes. The strict dependency is caused by https://github.com/mvukov/rules_ros2/blob/92338a842b0ad959211419438d558474f9d4f2fa/ros2/BUILD.bazel#L94-L106 and https://github.com/mvukov/rules_ros2/blob/92338a842b0ad959211419438d558474f9d4f2fa/ros2/interfaces.bzl#L829-L835 which in turn depends on the requirements_lock.txt and current_py_cc_headers version to be in sync.

.github/workflows/main.yml
lalten: removed

examples/WORKSPACE
lalten: rephrased.

examples/MODULE.bazel
mvukov: Shall we wait for https://github.com/bazelbuild/rules_rust/pull/3231 to be merged and a new rules_rust release? That way we can simplify this PR and make the documentation more streamlined.

mvukov:
@ahans I am curious how is

```
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
use_repo(pip, "rules_ros2_pip_deps")
```
useful for you. AFAIK, one cannot do
```
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(
    hub_name = "rules_ros2_pip_deps",
    python_version = "3.10",
    requirements_lock = "@rules_ros2//:requirements_lock.txt",
)
use_repo(pip, "rules_ros2_pip_deps")
```
So, how the former declaration helps you?

ahans:
> @ahans I am curious how is
> 
> ```
> pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
> use_repo(pip, "rules_ros2_pip_deps")
> ```
> 
> useful for you.

I use it exactly as outlined in my previous comment. Through those two lines I make `rules_ros2_pip_deps` available in my main module. Then I have a `requirement` macro that I use instead of the one from `rules_ros2_pip_deps` or `main_repo_pip_deps`. I also maintain a list with packages that are defined in `rules_ros2_pip_deps`. Then my macro forwards to the `requirement` from `rules_ros2_pip_deps` if a package is requested that exists in there. Otherwise, it goes to `main_repo_pip_deps`. If I don't do that, I might end up with duplicate packages. For `py_binary` from `rules_python`, that is not an issue (apparently they just pick the first one, as discussed previously here), but the rules from `rules_py` don't like that.

> AFAIK, one cannot do
> 
> ```
> pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
> pip.parse(
>     hub_name = "rules_ros2_pip_deps",
>     python_version = "3.10",
>     requirements_lock = "//:my_requirements_lock.txt",
> )
> use_repo(pip, "rules_ros2_pip_deps")
> ```

Making PyPI packages from the main repo available to `rules_ros2` would be my favorite solution. Then I would just make sure that all required packages are present and patch `rules_ros2` to use my repo (or live with the somewhat confusing fact that Python packages always come from `rules_ros2_pip_deps`). I haven't looked further into that yet, though. What I could do for sure to avoid the duplication issue is patching rules_ros2's `requirements_lock.txt` to also include all my packages and then just use that from the main repo as well. Above it sounds like that's the route most folks choose. So for that this pattern of pulling in a PyPI repo from another module is also useful.

examples/MODULE.bazel
mering: https://github.com/bazelbuild/rules_rust/pull/3231 is merged now.

lalten: Only need a new release now

lalten: Oh, looks like 0.58.0 is released actually! I can update when I find some time in the coming days

lalten: updated, bzlmod tests now use the same host clang toolchain as workspace. CI now tests everything for Bzlmod, including rust tests.

mvukov:
The impl is fine, I just need to find time to update documentation with both approaches and list caveats.

.bazelrc
```diff
@@ -49,12 +41,16 @@ test --build_tests_only
...
 test --test_output=errors
+build --verbose_failures
 
 # To use a clang compiler, invoke Bazel with `--config=clang`.
 build:clang --repo_env=CC=clang
 build:clang --repo_env=CXX=clang++
 build:clang --linkopt="-fuse-ld=lld"
 
+# Work around https://github.com/bazelbuild/rules_rust/issues/2125
+build --action_env=CARGO_BAZEL_REPIN=1
```
mvukov, line 52: Is this bzlmod related? I don't observe this on the main branch when I test examples (locally).

.bazelrc
```diff
@@ -49,12 +41,16 @@ test --build_tests_only
 
 # Show all the problems
 test --test_output=errors
+build --verbose_failures
```
mvukov, line 44: IMO, not related to this PR and can be removed.

.github/workflows/main.yml
```diff
@@ -91,13 +99,18 @@ jobs:
           path: |
             ~/.cache/bazel
             ~/.cache/bazel-repo
-          key: v4-bazel-repo-cache-${{ matrix.toolchain }}-${{ hashFiles('.bazelversion', 'examples/WORKSPACE', 'repositories/**', 'requirements_lock.txt', 'WORKSPACE') }}
-          restore-keys: v4-bazel-repo-cache-${{ matrix.toolchain }}-
-      - name: bazel test ${{ matrix.config_option }} //...
```
mvukov, line 96: It helps me to visually distinguish builds by having config_option in the name. I'd keep this one.

lalten:
It not optimal to try and keep this branch up to date with the ever-changing main. I've done my best to try and split off PRs to land things in smaller slices.
There will always be opportunities to improve docs or fix bugs but it shouldn't be done in this long-living branch where people can't even contribute with own PRs.
@mvukov what's your plan on merging this or the connected PRs?

.github/workflows/main.yml
lalten: updated

.bazelrc
lalten: removed

.bazelrc
lalten: This is indeed required because the lockfile had different shas in Workspace and Bzlmod, see the linked issue

examples/README.md
```diff
@@ -1,11 +1,48 @@
...
+
+```bash
+common --noenable_bzlmod --enable_workspace
+```
+
+### [bzlmod](https://bazel.build/versions/7.5.0/external/overview#bzlmod)-based setup
+
+This is the new external dependency subsystem. Please take a look at the `MODULE.bazel` file in this folder to get an idea how to set up yourself.
+
+With this approach, `rules_ros2` defines a Python 3.10 interpreter and defines the hub with resolved Python deps. If you want a different intepreter and/or you want to extend the Python requirements files with extra deps you might want to use in your monorepo, you'll need to patch `requirements_lock.txt` file and this repo.
```
otiv-milan, line 35: @lalten @ahans I added some docs about setup. Please review and tell me if something is missing. Would be good if you can list what needs to be patched if folks want to use additional deps in their repos.

mering: Could this be done in a follow-up in order to avoid some more weeks of back and forth? To me the current state seems already be good enough to use by many people. Without merging and publishing to BCR, it can't be used by anyone.

otiv-milan (approved):

mvukov (approved):

wep21:
Congrats!! 🎉

PR #245, opened 2024-01-15
Don't build netifaces wheel on pip parse

lalten:
Adds https://github.com/ros2/ros2cli/pull/875 as a patch.

[netifaces](https://github.com/al45tair/netifaces) is unmaintained: https://github.com/al45tair/netifaces/issues/78.
It does not provide official prebuilt wheels for Python > 3.9: https://pypi.org/project/netifaces/#files

We don't have a non-bazel host compiler so building that Python wheel is difficult in our CI.


repositories/patches/ros2cli_replace-netifaces.patch
```diff
@@ -0,0 +1,74 @@
...
+--- ros2cli/package.xml
++++ ros2cli/package.xml
+@@ -18,9 +18,9 @@
+ 
+   <exec_depend>python3-argcomplete</exec_depend>
+   <exec_depend>python3-importlib-metadata</exec_depend>
+-  <exec_depend>python3-netifaces</exec_depend>
+   <exec_depend>python3-packaging</exec_depend>
+   <exec_depend>python3-pkg-resources</exec_depend>
++  <exec_depend>python3-psutil</exec_depend>
```
mvukov, line 12: Oh, well, we don't need this, but I guess you wanted to recycle the patch from https://github.com/ros2/ros2cli/pull/875.

mvukov (approved):
LGTM, though I would've chosen ifaddr as it's a pure python impl. I guess you wanted to be in sync with from https://github.com/ros2/ros2cli/pull/875.

mvukov:
btw, psutil binaries are only for amd64. I guess you don't need e.g. arm64 at the moment?

PR #251, opened 2024-01-25
Drop dependency on Boost

lalten:
As pointed out in https://github.com/mvukov/rules_ros2/pull/238#issuecomment-1909969580, Boost is only needed for Foxglove. It turns out it isn't actually needed there either (anymore?).

In Foxglove itself, there is only a ROS 1 Smoketest that needs boost::filesystem. So nothing we need to care about. https://github.com/search?q=repo%3Afoxglove%2Fros-foxglove-bridge%20boost&type=code

In ROS 2's resource_retriever, It seems like there was some Boost going on in the past, but on the Humble branch there definitely isn't any: https://github.com/ros/resource_retriever/blob/humble/resource_retriever/include/resource_retriever/retriever.hpp

Unless I missed something it seems to be we can just drop the dependency on Boost altogether in rules_ros2.

### DIFF
--- a/repositories/deps.bzl
+++ b/repositories/deps.bzl
@@ -2,7 +2,6 @@
 """
 
 load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
-load("@com_github_nelhage_rules_boost//:boost/boost.bzl", "boost_deps")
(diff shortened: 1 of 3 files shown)


mvukov (approved):
Great, many thanks! :rocket:

PR #254, opened 2024-01-28
refactor: extract rules_ros2_workspace_deps

lalten:
Extracted from / in preparation for https://github.com/mvukov/rules_ros2/pull/238

No functional or version changes in this one. Only the WORKSPACE incantation changes.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,8 +1,8 @@
 workspace(name = "com_github_mvukov_rules_ros2")
 
-load("//repositories:repositories.bzl", "ros2_repositories", "rules_ros2_workspace_deps")
+load("//repositories:repositories.bzl", "ros2_repositories", "ros2_workspace_repositories")
 
-rules_ros2_workspace_deps()
+ros2_workspace_repositories()
 
 ros2_repositories()
 
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,6 +1,8 @@
 workspace(name = "com_github_mvukov_rules_ros2")
 
-load("//repositories:repositories.bzl", "ros2_repositories")
+load("//repositories:repositories.bzl", "ros2_repositories", "rules_ros2_workspace_deps")
+
+rules_ros2_workspace_deps()
 
 ros2_repositories()
 
--- a/repositories/repositories.bzl
+++ b/repositories/repositories.bzl
@@ -1,7 +1,6 @@
 """Handles imports of external/third-party repositories.
 """
 
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 load(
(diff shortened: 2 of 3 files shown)


WORKSPACE
```diff
@@ -1,6 +1,8 @@
 workspace(name = "com_github_mvukov_rules_ros2")
 
-load("//repositories:repositories.bzl", "ros2_repositories")
+load("//repositories:repositories.bzl", "ros2_repositories", "rules_ros2_workspace_deps")
+
+rules_ros2_workspace_deps()
```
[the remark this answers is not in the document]
lalten: Well yes that's possible, but it would have to be changed back for Bzlmod -- #238 continues to use `ros_repositories()` for ROS 2 repos, but all the other repos like googletest, asio, websocketpp, ... should be comoing from bzlmod directly.
I think it makes sense to have them separate.

If we really want to keep full backwards compatibility and not require WORKSPACE users to update which setup method is used, we can keep the `ros2_repositories` name and split out the ROS 2 repos into another macro that will be called directly by the Bzlmod extension. Not a huge fan of this option though because I think the `ros2_repositories` name would be a misleading.

mvukov:
I'd move console_bridge,  foxglove_bridge and probably also mcap to ros2_repositories() as those are ros-related. WDYT?

repositories/repositories.bzl
```diff
@@ -1,17 +1,15 @@
...
 load(
     "@com_github_mvukov_rules_ros2//repositories:ros2_repositories_impl.bzl",
     "ros2_repositories_impl",
 )
 
-def ros2_repositories():
-    """Imports external/third-party repositories.
-    """
+def rules_ros2_workspace_deps():
+    """Import http_archive dependencies for the WORKSPACE version of rules_ros2."""
```
mvukov, line 12:
```
"""Imports dependent third-party repositories for the non-blzmod (hence, workspace-) version of the repository.

In particular, imports third-party package repositories excluding ROS 2 packages. ROS 2-specific repositories are imported with `ros2_repositories()` macro.
"""

File: repositories/repositories.bzl [hunk 3]
```diff
@@ -1,17 +1,15 @@
...
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 load(
     "@com_github_mvukov_rules_ros2//repositories:ros2_repositories_impl.bzl",
     "ros2_repositories_impl",
 )
 
-def ros2_repositories():
-    """Imports external/third-party repositories.
-    """
+def rules_ros2_workspace_deps():
```
mvukov, line 11: Well, let's be consistent. I'd name this e.g. ros2_workspace_repositories(). IMO, this is still a bit ambiguous compared to ros_repositories, but the docs should do the explanation. IMHO, the suffix should be _repositories, not _deps. We use _deps for configuration of repositories.

Another alternative is to keep this as ros2_repositories and rename the other one as ros2_package_repositories() (which could be more specific).

WDYT?

lalten: Don't care too much, the best move would be to migrate to Bzlmod and then just drop Workspace altogether :D

PR #260, opened 2024-01-30
Bzlmod CI

lalten:
Another extraction from #238.
Obviously there isn't much Bzlmod going on on main right now, but we can already add the matrix dimension for it.

Note that the "required" checks in main branch protection rules have to be updated for this to land


mvukov:
IMO, this should be a part of the PR that introduces bzlmod.

PR #268, opened 2024-02-03
Simplify CI setup

lalten:
Let me know what you think about this one. I think having the actions/cache doesn't really make sense: It doesn't matter if a freshly spun up runner downloads the Bazel output base / Bazel action cache from GitHub's Cache or from BuildBuddy's Cache. In fact the action/cache will just blindly download the entire cache every time, even if the Bazel remote cache has all build and test results fully cached and nothing needs to be downloaded.
Same thinking for the Bazelisk caching. Doesn't matter if it's downloaded from the GH cache or from the Bazelisk download URL.

Additionally I cleaned up the CI test setup:
* Move ci.bazelrc to the repo root and symlink it in the same way as the .bazelrc in /examples.
* import the ci.bazelrc from .bazelrc, so there is no need to add --bazelrc options to the bazel invocation in CI
* Remove the test.sh and instead add --config=ci and the BB api key directly in the GHA yaml.

I believe a fully-cached build isn't possible in either case because of the many rules_foreign_cc CMake builds, which don't cache properly.

There is one gotcha: PRs from forks do not have access to the repo secrets, including the Buildbuddy api key. My suggestion is to create a [read-only key](https://www.buildbuddy.io/docs/guide-auth/#read-only-keys) at https://app.buildbuddy.io/settings/org/api-keys and then add that publicly visible in the .bazelrc (I put a TODO there as a placeholder for now). Builds will then use the read-only key by default. But in rules_ros2 repo GitHub Action runs it will replace the line in the .bazelrc the read/write key.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -42,6 +42,9 @@ build:clang --repo_env=CC=clang
 build:clang --repo_env=CXX=clang++
 build:clang --linkopt="-fuse-ld=lld"
 
+# Load --config=ci settings
+import %workspace%/ci.bazelrc
+
 # Load any settings specific to the current user.
 # user.bazelrc should appear in .gitignore so that settings are not shared with
 # team members. This needs to be last statement in this config,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,30 +72,14 @@ jobs:
         with:
           ref: ${{github.event.pull_request.head.sha}}
           repository: ${{github.event.pull_request.head.repo.full_name}}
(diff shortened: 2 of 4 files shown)


mvukov:
So, buildbuddy "bills" their users per amount of downloaded cache. I am not so keen to allow for a readonly token that everyone can use and then drain my free credits :) (and free credits are per user, hence for all my repos) If you want, I can give you write access to this repo such that you don't waste much time waiting for CI.

Besides remote config, what else bothers you about the CI config?

PR #299, opened 2024-04-03
C/C++: treat typesupport identifier libs as dynamic

mvukov:
An alternative and simpler approach to nail problems with type-support identifiers problems. 

@brians-neptune PTAL.
@mikebauer PTAL and tell me whether this works with your clang+libc++ combo.


mvukov:
@ahans PTAL as well and tell me what you think.

PR #341, opened 2024-06-29
Massage the bazelrc

lalten:


.github/workflows/ci.bazelrc
```diff
@@ -22,9 +22,7 @@ build:remote --noslim_profile
 
 # Don't spam CI logs.
 build --show_progress_rate_limit=60
+common --color=yes
 
-# Don't rely on test logs being easily accessible from the test runner,
-# though it makes the log noisier.
-test --test_output=errors
```
mvukov, line 28: I think this is fine for CI.

.bazelrc
```diff
@@ -37,6 +41,11 @@ build --noexperimental_check_output_files
 # Don't bother building targets which aren't dependencies of the tests.
 test --build_tests_only
 
+# Show all the problems
+test --keep_going
+test --test_output=errors
+build --verbose_failures
```
mvukov, line 47: I'm fine with `--test_output=errors` being here but the other two I don't see necessary by default (e.g. I use them on occasions, only).

.bazelrc
```diff
@@ -16,6 +16,10 @@ build --noenable_bzlmod
 # This is a mandatory flag.
 build --incompatible_default_to_explicit_init_py
 
+# Don't let local Python site packages leak into the build and cause problems
+common --incompatible_strict_action_env
```
mvukov, line 20: This one is already in this file, perhaps change that one from build to common.

.bazelrc
```diff
@@ -16,6 +16,10 @@ build --noenable_bzlmod
...
+common --action_env=PYTHONNOUSERSITE=1
```
mvukov, line 21: Very nice, thanks! :)

.bazelrc
lalten: removed

.github/workflows/ci.bazelrc
lalten: removed

.bazelrc
lalten: changed

mvukov (approved):

PR #343, opened 2024-06-29
Change `asio` BUILD to be compatible with the BCR version

lalten:

Change `asio` BUILD to be compatible with the BCR version:
https://github.com/bazelbuild/bazel-central-registry/blob/b58c330b8386c64d38c3587c4a1c55ef344fce14/modules/asio/1.28.2/patches/add_build_file.patch

Note that foxglove_bridge_base was previously transitively depending on `@boringssl//:ssl` through `asio`.


repositories/asio.BUILD.bazel
```diff
@@ -5,15 +5,10 @@ load("@rules_cc//cc:defs.bzl", "cc_library")
...
-    srcs = glob(["src/*.cpp"]),
-    hdrs = glob([
-        "include/**/*.hpp",
-        "include/**/*.ipp",
+    srcs = glob([
+        "**/*.hpp",
+        "**/*.ipp",
     ]),
-    defines = [
-        "ASIO_SEPARATE_COMPILATION",
```
mvukov, line 14: What's improvement here?

lalten: It's arguably worse to have asio as header only lib actually. But that's the way it's configured in the Bazel Central Registry. So to make sure things will work the same with Bzlmod and Workspace we need to change it to be compatible as we can't change the BCR one (at least it seems like more effort)

udaya2899: FWIW, you can create a new version in BCR the way you want, and if the module maintainer agrees, you can have a separate version like 1.28.1.bcr.1. That's what I did for osqp [here](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/osqp). I wanted a specific flag to be set differently, and I created a 0.6.3.bcr.1

Just saying, so you have the option to change at source.

mvukov: I'd vote we keep the impl from this repo as it's better, and upgrade the bzl one once that's fixed.

lalten: Updated the BUILD to match https://github.com/bazelbuild/bazel-central-registry/pull/2675/files#diff-d92e7004420fb61b89e25389540f8e1014cdde782379b35bad27adbd04f9e2c9
which in turn is based on this one.

mvukov (approved):

PR #165, opened 2023-08-15
Support android

ziye-dm:
Use the same configs with QNX and it's working.

PR #372, opened 2024-09-06
fix CMAKE_SYSTEM_NAME for Android and QNX

getim:
#165 added support for Android, and correctly combined a lot of settings with the QNX platform. However, it also started passing `CMAKE_SYSTEM_NAME=ANDROID_OR_QNX` to CMake, which is not a supported value. Instead pass `Android` or `QNX` depending on the current platform (see docs [here](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html)).

I had to re-organize a few config settings and some groups of flags passed together, but functionally the `CMAKE_SYSTEM_NAME` is the only value that's now passed different for the `Android` and `QNX` platform.

### DIFF
--- a/repositories/cyclonedds.BUILD.bazel
+++ b/repositories/cyclonedds.BUILD.bazel
@@ -40,17 +40,17 @@ selects.config_setting_group(
 )
 
 selects.config_setting_group(
-    name = "android_or_qnx",
-    match_any = [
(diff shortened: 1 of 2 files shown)


mvukov (approved):
Thanks and sorry for the late review.

PR #388, opened 2024-10-10
Clean up testing infra

mvukov:
Turns out `--sandbox_default_allow_network=false` does all necessary job and we don't need domain_id shenanigans.

@ahans @lalten PTAL.


lalten (approved):
lgtm. We already set --sandbox_default_allow_network=false on our side.

ros2/pytest_wrapper.py.tpl
```diff
@@ -41,28 +39,22 @@ def main() -> None:
     if 'ROS_LOG_DIR' not in os.environ:
         os.environ['ROS_LOG_DIR'] = bazel_test_output_dir
 
-    with contextlib.ExitStack() as stack:
```
ahans, line 44: I would keep this as an extra line of defense. I also find it good practice to use tempfile and not rely on Bazel for stuff like this. But if you really like it, we can always patch locally.

ahans (approved):
The changes are fine, but I would keep the local domain for testing to not mandate this Bazel setting from downstream projects. Also I'm not sure what'd happen if you bazel run a test. With the local domain it'll always work....

mvukov:
It turns out that ros2_cpp_test and ros2_py_test work correctly in the main branch only because `--sandbox_default_allow_network=false` is set -- otherwise there would be no network isolation.

I added additional docs to emphasize that folks should use network isolation. I think we are fine. If someone comes up with a valid test case where they need to get out of network isolation, we can act accordingly.

mvukov:
Another point: there are use-cases where ros2_test is just an overkill -- e.g. you want to unit test one node. e.g. ros2_cpp_test does the job in those cases. Moreover, ros2_test introduces Python dep and runtime overhead -- which annoys me sometimes...

PR #399, opened 2024-11-29
Disable warnings for cyclonedds dependency

mering:
Unforatuntely, there is no generic way to disable such warnings (https://github.com/bazel-contrib/rules_foreign_cc/issues/1332) but instead it needs to be done for every `rules_foreign_cc` target explicitly.
I also created https://github.com/eclipse-cyclonedds/cyclonedds/issues/2146 upstream to fix the warning in the first place such that this workaround can be removed in the future.


repositories/cyclonedds.BUILD.bazel
```diff
@@ -179,6 +179,10 @@ cmake(
         },
         no_match_error = "Unsupported build configuration",
     ),
+    copts = [
+        "-Wno-documentation-deprecated-sync",
+        "-Wno-error",
```
mvukov, line 184: Which compiler complains about this? Shall we just set `-w`?

mering: LLVM 19, I am fine with `-w` as well.

mvukov: OK, let's go with "-w" then.

mering: Done.

mvukov (approved):

PR #420, opened 2025-01-23
Set `rosidl_typesupport_introspection_{c,cpp}_identifier` target visibility `public`

udaya2899:
### Why?
A `cc_binary` we have that transitively depends on `ros2_rosidl` and when ran inside an oci_container, fails with 

```
error while loading shared libraries: librosidl_typesupport_introspection_c_identifier.so: cannot open shared object file: No such file or directory
```
To add the expected `.so` files to the container as a layer, we need the target that directly builds these `.so` files. For this, we need the visibility to be public.

PS: Thanks in advance for your work in maintaining this. Btw, looking forward for Bzlmod support soon!


mvukov:
Can you create a minimal example for this? I run ros2_launch apps inside docker containers produced with rules_oci without problems. I would expect no problems with ros2_cpp_binary targets. The problem you have may be that your pwd inside container isn't correct. I don't see how visibility would solve this.

udaya2899:
I'm hypothesizing this is because of using rules_ros2 as a `non_module_dep` in bzlmod currently using extension. 

>  I don't see how visibility would solve this.

To answer your question, setting it to public allows me to explicitly get the `.so` file and do:

```
container_layer(
    name = "rosidl_dependency_layer",
    # remap to /ros2_rosidl directory
    data_path = "/external/my_repo_name++extension_for_rules_ros2_ext+ros2_rosidl",
    directory = "/ros2_rosidl",
    files = [
        "@ros2_rosidl//:rosidl_typesupport_introspection_c_identifier",
        "@ros2_rosidl//:rosidl_typesupport_introspection_cpp_identifier",
    ],
)
```

and use this layer in `oci_image` as an extra layer so that the binary can find the `.so` files. Doing this works for us as the targets need to be public to build the `.so` files and add them. 

I will try to see if this is resolved by #238 before creating a minimal repro repository for this.

mvukov:
BTW, just to ask here: what do you pack in an image: a C++ target or a ros2_launch target?

mvukov:
Can you check that if that extra layer creates duplicates of the .so files (you can use dive utility for that)? If so, it's probable that you don't start your app inside a correct dir.

udaya2899:
No they are symlinks. I checked with dive and the layers I add are basically symlinking to the public .so files (that are made public through this PR).

I cannot depend on these .so files and create symlinks from HEAD.

I cannot use the publicly exposed target for this since the .so file is not an output there.

mvukov (approved):

mvukov:
@udaya2899 Looks like you have a custom (linking) setup. The changes LGTM. But, for the future reference: packaging ros2_launch targets works just fine without the aforementioned hacks.

udaya2899:
Thanks for the approval. Seems the CI checks still need some approval from you to run

mering:
@mvukov @otiv-milan Could you approve the workflow again?

PR #424, opened 2025-02-09
Fix versions for empy, numpy and types-pkg-resources

mvukov:


ryan-brigden-ai:
@mvukov I've done some digging to understand why numpy is being pinned here to `1.23.*`. Any chance you could briefly elaborate on why this constraint is necessary? I'd be interested in contributing to allow for support of modern `numpy` versions (2.X of interest).

mvukov:
Hi, there is no hard reason except to make sure numpy headers are correctly found in https://github.com/mvukov/rules_ros2/blob/main/ros2/BUILD.bazel#L94-L105. You can try to update numpy to a recent version and possibly fix the path the headers.

PR #427, opened 2025-02-10
Add Python transition attrs to _ROS2_COLLECTOR_ATTR_ASPECTS

lalten:

It looks like things have changed in https://github.com/bazelbuild/rules_python/commit/611eda87e48a3943808216cf4e4de875040b09aa but it might be nice to keep this change to stay compatible for users who need to patch-in older rules_python.

### DIFF
--- a/ros2/plugin_aspects.bzl
+++ b/ros2/plugin_aspects.bzl
@@ -38,15 +38,16 @@ Ros2PluginCollectorAspectInfo = provider(
     ],
 )
 
-_ROS2_COLLECTOR_ATTR_ASPECTS = ["data", "deps"]
+# From https://github.com/bazelbuild/rules_python/blob/0.33.2/python/config_settings/transition.bzl#L121-L140
+_py_transition_attrs = ["target", "tools"]
+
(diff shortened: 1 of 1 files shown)


ros2/plugin_aspects.bzl
```diff
@@ -38,15 +38,16 @@ Ros2PluginCollectorAspectInfo = provider(
     ],
 )
 
-_ROS2_COLLECTOR_ATTR_ASPECTS = ["data", "deps"]
+# From https://github.com/bazelbuild/rules_python/blob/0.33.2/python/config_settings/transition.bzl#L121-L140
+_py_transition_attrs = ["target", "tools"]
```
mvukov, line 42: Do I understand this correctly that this is only needed for older rules_python than 1.x?

lalten: The tag in the comment was the latest one at the time of writing. Later rules_python restructured that file so I don't immediately find the line on the current latest tag

mvukov: TBH, I've never really given too many thoughts about backwards compatibility. The project, at the moment, follows live-at-head philosophy and implicitly that also means that users of that repo have to accept that.

mvukov:
IMO, no need for this PR.

PR #430, opened 2025-02-14
Misc fixes

lalten:
Here are three patches that we use locally which would be nice to upstream.


repositories/cyclonedds.BUILD.bazel
```diff
@@ -181,6 +181,7 @@ cmake(
     ),
     copts = [
         "-w",
+        "-D__FILE__=\"redacted\"",
```
mvukov, line 184: Is this determinism related? Doesn't Bazel already set this via compiler flags?

ros2/cc_defs.bzl
```diff
@@ -75,7 +75,7 @@ def _ros2_cpp_exec(target, name, ros2_package_name, set_up_ament, idl_deps, **kw
 
     launcher_target_kwargs, binary_kwargs = split_kwargs(**kwargs)
     target_impl = name + "_impl"
-    _ros2_cc_target(cc_binary, "cpp", target_impl, ros2_package_name, tags = ["manual"], **binary_kwargs)
+    _ros2_cc_target(cc_binary, "cpp", target_impl, ros2_package_name, tags = launcher_target_kwargs.get("tags"), **binary_kwargs)
```
mvukov, line 78: What's the reasoning here?

repositories/cyclonedds.BUILD.bazel
lalten: Generally yes, but CMake involvement caused issue here that were solved by this

ros2/cc_defs.bzl
lalten: We use tags for internal tooling (e.g. doing some build graph analysis and restrictions). The rule was throwing away the tags and unconditionally replacing it with `manual`. This change keeps the tags for the cc_binary target. The manual one in L89 is still there.

repositories/cyclonedds.BUILD.bazel
mvukov: Just for my understanding, what was the issue?

ros2/cc_defs.bzl
mvukov: Hm, internal targets are typically marked as manual. The top level target should receive the tags passed through macro.  Do you need this for e.g. some code analysis tooling?

lalten: Yup, e.g. clang-tidy aspect integration.

repositories/cyclonedds.BUILD.bazel
lalten: absolute paths in binaries, which causes indeterministic builds

ros2/cc_defs.bzl
mvukov: I get your intention. Though, IMO, this is a very custom setup. FWIK, internal targets are best to be tagged as manual. Instead of relying on tags, you can query for target kind and then run analyzers on such targets.

mvukov (commented):
Left a small comment, otherwise LGTM.

ros2/cc_defs.bzl
lalten: that's not really the same -- for custom tools you in many cases surely you could go with queries instead. But some existing tools like https://github.com/erenon/bazel_clang_tidy rely on tags at this time.

mvukov: You refer to this thing https://github.com/erenon/bazel_clang_tidy/blob/master/example/BUILD#L17, right? I'm fine to set `launcher_target_kwargs.get("tags") + ["manual"]`. Does that work for you?

lalten: sgtm, updated.

mvukov (approved):

PR #432, opened 2025-02-14
Fix  empty globs and rclcpp_components

### LINKED ISSUE #431: Only a fraction of repos is built in CI
Bazel 8.0.0 flips `--incompatible_disallow_empty_glob` to true. We have quite a few empty globs. I was about to prepare a PR that fixes them for `bazel test //...` in the root of the repo, when I figured I should also test in `examples`. Guess what, that found more issues. And then I realized that we actually don't even check if everything builds. We should change that, shouldn't we? I'd propose to have a meta `cc_library` as well as a `py_library` in `ros2/test/BUILD.bazel` that depends on all `cc_` and `py_` targets, respectively. A `bazel test //...` would then also build that. It'd be probably ok to maintain those targets manually. Alternatively, something based on `bazel query '//external:*' ` could probably work as well. Wdyt?

ahans:
Fixes #431

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -16,6 +16,9 @@ build --noenable_bzlmod
 # This is a mandatory flag.
 build --incompatible_default_to_explicit_init_py
 
+# This is flipped in 8.0.0, so better make sure we're compatible already now.
+build --incompatible_disallow_empty_glob
+
 # All code will be compiled with C++17 flag.
 # Users can naturally override copts for cc_* targets.
 # NOTE: googletest >=v1.13.0 requires min C++14.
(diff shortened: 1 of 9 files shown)


ahans:
I think the build fixes are all valid, but the CI change could use some improvement, as now the external repos are built twice, once from root and once from `examples`.

ahans:
Some things seem to be off. But you get the idea.

mvukov:
Fixed in #433 and #435.

mvukov:
Your PR has more fixes, but we don't have unit tests for them. Those remaining fixes should come with some unit tests and there should be no changes to the CI IMO.

PR #445, opened 2025-02-25
Add rosbag2 keyboard handler patch

lalten:
This removes the registering a keyboard handler from the system when using the rosbag2 keyboard handler.

With this patch, you can set nullptr as keyboard handler and it will not fiddle with the system internals. If not supplied, it will work as before...

Background is that this causes side effects on the system:

- Terminal will ignore any further inputs and freeze.
- Tests will not properly finish and crash on gtest cleanup code.


mvukov:
Just curious: is this patched in the official repo as well?

lalten:
> Just curious: is this patched in the official repo as well?

There is indeed something similar in https://github.com/ros2/rosbag2/commit/4882d30fc2d1e5b9305e5d46b8460466f9280d27 but that's not backported to Humble. The patch in this PR is much simpler and has the same effect.

lalten:
@mvukov should we land #443 #444 #445 and #446?

PR #446, opened 2025-02-25
Fix RMW_LIBRARY_PATH define in presence of transitions

lalten:
This fixes the below error
```
unknown file: Failure
C++ exception with description "Failed to deserialize ROS message.: failed to load shared library 'external/ros2_rmw_cyclonedds/librmw_cyclonedds.so' due to dlopen error: external/ros2_rmw_cyclonedds/librmw_cyclonedds.so: cannot open shared object file: No such file or directory, at external/ros2_rcutils/src/shared_library.c:96, at external/ros2_rmw_implementation/rmw_implementation/src/functions.cpp:90" thrown in the test body.
```

Which you will get when transitions (e.g. via with_cfg.bzl) are at play, because the `k8-fastbuild` and friends will not be correct anymore, it has to be `k8-fastbuild-ST-61f4e3852c4b` or similar, depending on configuration.

Using `../` instead is always correct because of the runfiles structure at runtime.

### DIFF
--- a/repositories/rmw_implementation.BUILD.bazel
+++ b/repositories/rmw_implementation.BUILD.bazel
@@ -15,9 +15,7 @@ ros2_cpp_library(
         "@ros2_rmw_cyclonedds//:rmw_cyclonedds",
     ],
     includes = ["include"],
-    local_defines = [
-        "RMW_LIBRARY_PATH=\\\"$(rootpath @ros2_rmw_cyclonedds//:rmw_cyclonedds)\\\"",
-    ],
+    local_defines = ['RMW_LIBRARY_PATH=\\"../ros2_rmw_cyclonedds/librmw_cyclonedds.so\\"'],
     visibility = ["//visibility:public"],
     deps = [
         "@ros2_rcpputils//:rcpputils",


mvukov:
`rootpath` shouldn't have `k8-fastbuild` in it AFAIK.

We touched this in
* https://github.com/mvukov/rules_ros2/pull/65 and then in
* https://github.com/mvukov/rules_ros2/pull/66

I'd like to learn more about your usecase. 

BTW, do you patch rmw_implementation with_cfg.bzl locally?

lalten:
I added a failing test in https://github.com/mvukov/rules_ros2/pull/446/commits/ca7da17e6b480bf79fdf24614f4afcc5475e7f87:
https://github.com/mvukov/rules_ros2/actions/runs/13606301154/job/38038169533?pr=446#step:5:61
image

and fixed it by using `../` instead of `$(rootpath` in 00f4065ceb624c82f1eb1519dc46f52af99b804c

repositories/repositories.bzl
```diff
@@ -131,6 +131,14 @@ def ros2_workspace_repositories():
         urls = ["https://github.com/zaphoyd/websocketpp/archive/refs/tags/0.8.2.tar.gz"],
     )
 
+    maybe(
+        http_archive,
+        name = "with_cfg.bzl",
+        sha256 = "d83f99ac39cd9940848ea11a51a60159cf09cda2ba30545036041551aae73ab4",
+        strip_prefix = "with_cfg.bzl-0.6.1",
+        url = "https://github.com/fmeum/with_cfg.bzl/releases/download/v0.6.1/with_cfg.bzl-v0.6.1.tar.gz",
```
lalten, line 139: yes this isn't the latest release, but newer ones seem to cause issues with struct members of `@bazel_features` not being available when building in Workspace mode

mvukov:
Would `--nolegacy_external_runfiles`  help in the first place (mentioned in #65 )? (that is a recommended flag, we should have it in .bazelrc eventually) . With that flag I think we should just get `../ros2_rmw_cyclonedds/librmw_cyclonedds.so`.

lalten:
> Would `--nolegacy_external_runfiles` help in the first place (mentioned in #65 )? (that is a recommended flag, we should have it in .bazelrc eventually) . With that flag I think we should just get `../ros2_rmw_cyclonedds/librmw_cyclonedds.so`.

True, this does indeed help. But requiring this to be set is a hard sell to downstream consumers who may have a lot of migration work ahead of them to support this flag being flipped -- especially when just using `../` also makes this work regardless of `--(no)legacy_external_runfiles`

mvukov:
We have a number of flags in .bazelrc without this repo can't work. I'm fine to declare `--nolegacy_external_runfiles` as a recommended one referencing this PR. `--nolegacy_external_runfiles` is a nice perf improvement  as it reduces the number of files in runfiles folder.

PR #456, opened 2025-03-01
Update to rules_rust(_bindgen) 58

lalten:

### DIFF
--- a/repositories/rust/rustfmt/generate.sh
+++ b/repositories/rust/rustfmt/generate.sh
@@ -2,4 +2,7 @@
 
 set -o errexit -o nounset
 
-bazel run @rules_rust//tools/upstream_wrapper:rustfmt --script_path="$(dirname "$0")/rustfmt_wrapper_impl.sh"
+bazel run \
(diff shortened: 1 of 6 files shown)


repositories/rust/rustfmt/generate.sh
```diff
@@ -2,4 +2,7 @@
 
 set -o errexit -o nounset
 
-bazel run @rules_rust//tools/upstream_wrapper:rustfmt --script_path="$(dirname "$0")/rustfmt_wrapper_impl.sh"
+bazel run \
+    --noenable_bzlmod --enable_workspace \
```
mvukov, line 6: Why is this change needed? Bazel should pick up bazelrc config.

lalten: It's from the Bzlmod PR but I thought I'd sneak it in here too as it has no adverse effects when there's no Bzlmod yet. The issue is that the script doesn't work with Bzlmod (but also it doesn't need to)

mvukov:
Closing in favor of #459.

PR #467, opened 2025-04-03
Add diagnostics monitor nodes

thomasegriffith:
Add diagnostics monitor nodes from `diagnostic_common_diagnostics`.

Includes the following updates:
- add bazel targets for `cpu_monitor_node`, `hd_monitor_node`, `ntp_monitor_node`, `ram_monitor_node`, and `sensors_monitor_node`
- update our version of the `ros/diagnostics` repo to `207f0c0f1f8e9bd1842a07d90459aa9a2b670c23`
- add `ntplib` to the pip requirements files


ros2/test/diagnostics/diagnostic_publisher.cc
```diff
@@ -20,7 +20,7 @@
 
 class Publisher : public rclcpp::Node {
  public:
-  Publisher() : Node("heartbeat_publisher"), diagnostic_updater_{this} {
+  Publisher() : Node("heartbeat_publisher"), diagnostic_updater_(this) {
```
mvukov, line 23: This change seems to be unrelated to this PR. Please revert.

thomasegriffith: The CI fails without this change.  Looks like the log [expired](https://github.com/mvukov/rules_ros2/actions/runs/14248663364/job/39935938834), but it's a brace initialization problem.

Between versions `9f402787ea2c9b3dd4d7e51a9986810e8a3400ba` and `207f0c0f1f8e9bd1842a07d90459aa9a2b670c23` of `ros/diagnostics`, they moved one of the `Updater` constructors from a header file to a source file.  That constructor is then used by the templated constructor.  Since we're using the templated constructor, brace initialization requires that its definition is completely visible from the header file.  Since it's delegating to another constructor that's now defined in the source file, brace initialization won't work.

I should have put the reason for this in the PR description.  Added!

thomasegriffith: To be more clear: they moved the definition of [this constructor](https://github.com/ros/diagnostics/blob/ros2/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp#L381-L388) to a source file.  

It's used by the templated constructor on [line 371](https://github.com/ros/diagnostics/blob/ros2/diagnostic_updater/include/diagnostic_updater/diagnostic_updater.hpp#L371).

repositories/diagnostics.BUILD.bazel
```diff
@@ -93,3 +96,71 @@ ros2_cpp_binary(
...
+    deps = [
+        ":py_diagnostic_updater",
+        "@ros2_common_interfaces//:py_diagnostic_msgs",
+        "@ros2_rclpy//:rclpy",
+        requirement("psutil"),
+    ],
+)
+
+ros2_py_binary(
+    name = "hd_monitor_node",
```
mvukov, line 115: What is `hd`?

repositories/diagnostics.BUILD.bazel
```diff
@@ -93,3 +96,71 @@ ros2_cpp_binary(
         "@ros2_rclcpp//:rclcpp",
     ],
 )
+
+ros2_py_binary(
```
mvukov, line 100: Sorry for the delay. Why do you need `set_up_ament = True`? How is this supposed to work in practice? Just launch the nodes?

repositories/diagnostics.BUILD.bazel
thomasegriffith: It stands for "hard drive".  `hd_monitor_node` has an unfortunate name, but I figure we ought to keep the naming consistent with the upstream stuff.  To help out on our end, I added a comment clarifying this is "hard drive".

repositories/diagnostics.BUILD.bazel
thomasegriffith: I suppose we don't need it.  Removed.  Thanks!

repositories/repositories.bzl
```diff
@@ -301,9 +301,9 @@ def ros2_repositories():
         http_archive,
         name = "ros2_diagnostics",
         build_file = "@com_github_mvukov_rules_ros2//repositories:diagnostics.BUILD.bazel",
-        sha256 = "a723dae7acf0f00ee643c076c7c81299be0254919f29225ec7a89dc14cb8ea6f",
-        strip_prefix = "diagnostics-9f402787ea2c9b3dd4d7e51a9986810e8a3400ba",
-        urls = ["https://github.com/ros/diagnostics/archive/9f402787ea2c9b3dd4d7e51a9986810e8a3400ba.zip"],
+        sha256 = "66b00679cc5404fbe41c4ee486382235f80900bc21bd7937ba25ffca469ec5d4",
+        strip_prefix = "diagnostics-207f0c0f1f8e9bd1842a07d90459aa9a2b670c23",
+        urls = ["https://github.com/ros/diagnostics/archive/207f0c0f1f8e9bd1842a07d90459aa9a2b670c23.zip"],
```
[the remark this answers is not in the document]
thomasegriffith: Good idea!  Done

repositories/diagnostics.BUILD.bazel
```diff
@@ -93,3 +96,73 @@ ros2_cpp_binary(
...
+    ],
+)
+
+# The "hd" in "hd_monitor_node" means "hard drive".  We name it "hd_monitor_node" to stay
+# consistent with the upstream name.
+ros2_py_binary(
+    name = "hd_monitor_node",
+    srcs = ["diagnostic_common_diagnostics/diagnostic_common_diagnostics/hd_monitor.py"],
+    main = "diagnostic_common_diagnostics/diagnostic_common_diagnostics/hd_monitor.py",
+    set_up_ament = False,
```
mvukov, line 120: Why do you need `set_up_ament = False` here and below? If not necessary, please remove.

mvukov (commented):
One small comment, otherwise LGTM.

PR #469, opened 2025-04-17
Add support for Python 3.11 with Bzlmod

mering:
Python 3.10 will continue to stay the default.

I tested compatibility by temporarily setting the default to 3.11 and running tests.


MODULE.bazel
```diff
@@ -44,28 +44,47 @@ http_archive(
...
-    hub_name = "rules_ros2_pip_deps",
-    python_version = _DEFAULT_PYTHON,
-    requirements_lock = "//:requirements_lock.txt",
-)
+
+[
+    pip.parse(
+        hub_name = "rules_ros2_pip_deps",
+        python_version = python_version,
+        requirements_lock = "//:requirements_lock.txt",
```
otiv-milan, line 70: The lock file is resolved for Pyhon 3.10 and may not work for other Python versions.

MODULE.bazel
```diff
@@ -44,28 +44,47 @@ http_archive(
...
-    hub_name = "rules_ros2_resolver_deps",
-    python_version = _DEFAULT_PYTHON,
-    requirements_lock = "//repositories/private:resolver_requirements_lock.txt",
-)
+
+[
+    dev_pip.parse(
+        hub_name = "rules_ros2_resolver_deps",
+        python_version = python_version,
+        requirements_lock = "//repositories/private:resolver_requirements_lock.txt",
```
otiv-milan, line 83: Same here.

MODULE.bazel
mering: Or do you prefer if I create a dedicated file for Python 3.11?

mvukov:
High level question (I am not that versed yet with bzlmod): how can folks use e.g. Python 3.11?

mvukov:
Also, what's the reasoning for 3.11 here? Why not 3.13 (which is the newest version at the moment)?

mering:
> High level question (I am not that versed yet with bzlmod): how can folks use e.g. Python 3.11?

The `default` attribute in a toolchain defined by `rules_python` in the root module has precedence.

Create `MODULE.bazel` with:
```starlark
module(name = "rules-ros2-python-example")

bazel_dep(name = "com_github_mvukov_rules_ros2")
git_override(
        module_name = "com_github_mvukov_rules_ros2",
        commit = "85abd3a47632a5c1a407feb2e249e1cdbd8dff41",
        remote = "https://github.com/mvukov/rules_ros2",
)

bazel_dep(name = "rules_python", version = "1.2.0")
python = use_extension("@rules_python//python/extensions:python.bzl", "python")
python.toolchain(
    is_default = True,
    python_version = "3.11",
)
```

Building this fails with
<details>
<summary>Log</summary>

```
$ bazel build @com_github_mvukov_rules_ros2//ros2:all
DEBUG: HOME/.cache/bazel/_bazel_USER/90a2ee47c8d01e4b8a59f404c13d3af0/external/rules_python+/python/private/config_settings.bzl:200:14: The current configuration rules_python config flags is:
    @@rules_python+//python/config_settings:pip_whl: "auto"
    @@rules_python+//python/config_settings:pip_whl_glibc_version: ""
    @@rules_python+//python/config_settings:pip_whl_muslc_version: ""
    @@rules_python+//python/config_settings:pip_whl_osx_arch: "arch"
    @@rules_python+//python/config_settings:pip_whl_osx_version: ""
    @@rules_python+//python/config_settings:py_freethreaded: "no"
    @@rules_python+//python/config_settings:py_linux_libc: "glibc"
    @@rules_python+//python/config_settings:python_version: "3.11"

If the value is missing, then the default value is being used, see documentation:
https://rules-python.readthedocs.io/en/latest/api/rules_python/python/config_settings
ERROR: /usr/local/google/home/schnirring/.cache/bazel/_bazel_schnirring/90a2ee47c8d01e4b8a59f404c13d3af0/external/rules_python++pip+rules_ros2_pip_deps/numpy/BUILD.bazel:5:12: configurable attribute "actual" in @@rules_python++pip+rules_ros2_pip_deps//numpy:_no_matching_repository doesn't match this configuration: No matching wheel for current configuration's Python version.

The current build configuration's Python version doesn't match any of the Python
wheels available for this distribution. This distribution supports the following Python
configuration settings:
    //_config:is_cp310

To determine the current configuration's Python version, run:
    `bazel config <config id>` (shown further below)

For the current configuration value see the debug message above that is
printing the current flag values. If you can't see the message, then re-run the
build to make it a failure instead by running the build with:
    --@@rules_python+//python/config_settings:current_config=fail

However, the command above will hide the `bazel config <config id>` message.


This instance of @@rules_python++pip+rules_ros2_pip_deps//numpy:_no_matching_repository has configuration identifier 3723722. To inspect its configuration, run: bazel config 3723722.

For more help, see https://bazel.build/docs/configurable-attributes#faq-select-choose-condition.

Target @@com_github_mvukov_rules_ros2+//ros2:rcl_logging_impl up-to-date (nothing to build)
```

</details>

Changing the commit to the commit from this PR makes the build succeed.

> Also, what's the reasoning for 3.11 here? Why not 3.13 (which is the newest version at the moment)?

The same lock file did not resolve correctly for the other versions. One probably needs to bump the versions for some of the dependencies such that there are compatible wheels available on pypi.org. This could come in a follow-up.

mvukov (approved):

PR #516, opened 2025-07-28
[WIP] Support testonly in `ros2_py_test()`

mikebauer:
Currently `ros2_py_test()` targets appear not to be able to depend on `testonly` targets. Propagate `testonly` into the inner `py_test` target correctly.

TODO(mikebauer) Figure out the right way to test this. The current way of testing this is to add an empty testonly `filegroup()` dependency to `//ros2/test/launch_pytest:ros2_py_test_rclpy_init_test` which fails without the `py_defs.bzl` changes.


mvukov:
Thanks for the fix. Please fix failing tests.

PR #554, opened 2025-10-15
Make launch.sh(.tpl) Posix shell compliant

lalten:
Make launch.sh Posix shell compliant.

In my case the `ctx.toolchains[SH_TOOLCHAIN].path` expands to `/bin/sh` which isn't bash but dash, which for example does not support `pipefail` but is fully Posix compliant


lalten:
I checked this again and there is no other change needed here

PR #588, opened 2026-01-07
Update deps and add load statements for Bazel 9 support

ahans:
Bazel 9 drops the builtin CC and Python rules. This commit adds missing load statements and updates upstream packages to fixed versions.

### DIFF
--- a/repositories/clang_configure.bzl
+++ b/repositories/clang_configure.bzl
@@ -17,6 +17,8 @@ def _clang_configure_impl(repository_ctx):
     repository_ctx.symlink(libclang_path, "libclang.so")
 
     repository_ctx.file("BUILD.bazel", """\
+load("@rules_cc//cc:defs.bzl", "cc_import")
+
 exports_files([
     "clang",
 ])
--- a/ros2/interfaces.bzl
+++ b/ros2/interfaces.bzl
@@ -16,6 +16,7 @@
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@com_github_mvukov_rules_ros2//ros2:cc_opts.bzl", "C_COPTS")
+load("@rules_cc//cc:defs.bzl", "CcInfo", "cc_common")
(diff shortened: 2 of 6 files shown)


.bazelversion
```diff
@@ -1 +1 @@
-8.5.0
+9.0.0rc3
```
ahans: Only here to have CI use Bazel 9. We would either wait until the final 9.0.0 is released and update to that before merging or merge without the change to `.bazelversion`.

ahans: Actually, if `WORKSPACE` support is supposed to be kept alive, we should not move to 9.0.0 as the only Bazel version this repo is tested with. It'd probably be best to extend the CI setup to use different Bazel versions (where with >= 9.0.0 we only test `bzlmod`). 9.0.0 removes bzlmod support. Tests still pass, because `--enable_workspace` is a no-op now (as is `--noenable_bzlmod`), so we effectively test the bzlmod version twice.

mvukov:
We'll need to work out CI config such that bzlmod is tested with bazel 9.x and workspace with 8.x.

ahans:
> Can we have load statements in this PR and the rest in a follow-up? Are there separate renovate in #129 already?

Sure, that would be sufficient for downstream Bazel 9 compliance, because people could just in their main module pick newer versions of the deps that currently use non-compliant versions.

For CI I'll figure something out. Would also have to go in a follow-up, because we can only set that live once all deps are updated.

ahans:
@mvukov I updated the PR to only have the load statements. This is not sufficient for Bazel 9 in this repo alone, but should help users that upgrade versions themselves.

CI will come in a follow-up, but for that to work we'd also need the upgrades for the 3rd party deps. I'll check #129 and tag you on the once that are already matching.

mvukov (approved):

ahans:
by rebasing and pushing i disabled the auto merge, can you please merge manually @mvukov ?

Issue #590, opened 2026-01-15
Update websocketpp to 0.8.2.bcr.6 and publish another release to the BCR


kgreenek:
The latest version of websocketpp bumps its dependency on boost.asio to 1.89.0.bcr.2, which has a compatibility_level of 0. Previous versions of boost had different compatibility levels based on the minor version (e.g. 108900 for 1.89.0). That caused some headaches for module maintainers. For more context see: https://github.com/bazelbuild/bazel-central-registry/discussions/6511

websocketpp is the only module that indirectly depends on boost which is used by rules_ros2. By updating it, rules_ros2 can be used in the same repo as any boost version that compiles.

It would be great to update the websocketpp version and then publish another release to the BCR. I also think publishing another release to the BCR would be a good idea anyways before the Jazzy PR lands. Happy to be the one to do that.

Thank you!

PR #591, opened 2026-01-15
Update websocketpp to 0.8.2.bcr.6

kgreenek:
Closes https://github.com/mvukov/rules_ros2/issues/590

This version depends on the latest bcr version of boost, which reset the compatibility_level to 0. This is the only indirect dependency on boost in rules_ros2 that I am aware of.


kgreenek:
I'm seeing test failures that look unrelated to my change. I'll try kicking again to see if I can get it to pass

PR #600, opened 2026-01-30
Upgrade to Numpy 2; Drop Python 3.10

lalten:
Python 3.13+ as well as most newer libraries don't support Numpy 1.x anymore, so it's time to upgrade.

At the same time, Python 3.10 doesn't support Numpy 2.x
But Python 3.10 is EOL in 2026-10 so maybe it's ok to drop that now.

See https://github.com/mvukov/rules_ros2/pull/599
See https://github.com/mvukov/rules_ros2/pull/424


mvukov:
Thanks for the PR. The reason for 3.10 was to be compatible with the default ROS 2 humble, IIRC. In #558 I upgraded already to Python 3.12 -- as that's official Python version for Jazzy. I plan to merge jazzy PR very soon. But I'll keep a yet-to-be-made humble branch alive for future patches; no worries about that.

I remember that the resolver deps needed to be upgraded in #558 , some additions were necessary.

FWIW, I'd vote we streamline the whole repo on a single Python version. Is it already safe to upgrade everything to Python 3.14? e.g. all common/ML packages exist for 3.14 already.  Haven't tested myself TBH.

mvukov:
Moved to Python 3.12 already. We can move to 3.14, why not.
Regarding numpy, still OK to upgrade it.

PR #604, opened 2026-02-05
Use Bazel 9.1.0 for bzlmod in CI

ahans:
Here's a proposal for CI integration. It is based on my other branch, whose one commit you can ignore here. I cherry-picked the three renovate commits (I tagged you on the PRs), looks like those are actually already sufficient (at least locally I got a green build), so no extra upgrades necessary for Bazel 9!


ahans:
@mvukov I rebased this branch, things are looking good now. What do you think? Would be good to ensure Bazel 9 compatibility, it is currently there, with the CI check we'd make sure it stays this way.

mvukov:
Sure this, many thanks.

PR #617, opened 2026-03-25
Add support for message definitions in omg idl syntax

woutervdstoel:
This PR adds support for directly including .idl files as message definitions rather than having them be generated by rosidl. It also supports mixing and matching of idl and msg definitions.

In our repo I also verified this works with rosbag recording and replay.


ros2/plugin_aspects.bzl
```diff
@@ -99,11 +99,12 @@ def create_interface_struct(target):
     return struct(
         package_name = target.label.name,
         srcs = target[Ros2InterfaceInfo].info.srcs,
+        generated_idl_files = target[IdlAdapterAspectInfo].generated_idl_files if IdlAdapterAspectInfo in target else None,
```
mvukov: Why are changes needed in this file?

mvukov (commented):
Last time I looked at this, there were constraints:
* 1 msg in .idl file -> ROS msg
* 2 msg in .idl file -> ROS srv
* 3 msg in .idl file -> ROS action
and that logic is/was pretty much hard-coded in the parser (IIRC).

There should be decent disclaimer about this. Instead of silently failing, what can be done to ensure proper error handline? -- not for this PR but I am curious.

Also, what's the motivation to use .idl directly?

ros2/interfaces.bzl
```diff
@@ -533,6 +573,11 @@ _TYPESUPPORT_INTROSPECION_GENERATOR_CPP_OUTPUT_MAPPING = [
 ]
 
 def _cpp_generator_aspect_impl(target, ctx):
+    if Ros2InterfaceInfo not in target:
```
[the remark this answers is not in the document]
woutervdstoel: True, not strictly related to this PR. I can create a separate PR for these aspect changes if you prefer.

In our setup, we want to apply this aspect transitively through some cc_library targets (that don't provide Ros2InterfaceInfo). This is similar to the old proto/grpc propagation through `deps` (which I suppose this aspect-based compilation was inspired by).

ros2/interfaces.bzl
```diff
@@ -693,9 +737,9 @@ def _py_generator_aspect_impl(target, ctx):
         "_{}_s.ep.{}.c".format(package_name, type_support_impl),
     ]
 
-    for ext in ["action", "msg", "srv"]:
-        if any([f.extension == ext for f in srcs]):
-            extra_generated_outputs.append("{}/__init__.py".format(ext))
+    for f in srcs:
+        if f.extension in ["action", "idl", "msg", "srv"]:
+            extra_generated_outputs.append("{}/__init__.py".format(_get_idl_submodule_name(f)))
```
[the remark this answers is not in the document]
woutervdstoel: To be honest I can't remember, removed.

ros2/plugin_aspects.bzl
woutervdstoel: I explained in a comment in `ament.bzl` why we also need the  `generated_idl_files` . I'm open to implementation suggestions that don't involve changing this struct because I agree this is a bit of a detour.

ros2/ament.bzl
```diff
@@ -200,17 +200,41 @@ def _ros2_ament_setup_rule_impl(ctx):
...
+                    paths.join(prefix_path, "share", package_name, "msg", src.basename),
+                )
+                ctx.actions.symlink(
+                    output = src_file,
+                    target_file = src,
+                )
+                outputs.append(src_file)
+                idl_manifest_contents.append(paths.join("msg", src.basename))
+
+            elif src.extension == "idl":
```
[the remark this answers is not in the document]
woutervdstoel: Updated the PR with simplified logic and some explanation about `generated_idl_files`.

ros2/plugin_aspects.bzl
```diff
@@ -99,11 +99,12 @@ def create_interface_struct(target):
...
     )
 
 def _ros2_interface_collector_aspect_impl(target, ctx):
     direct_interfaces = []
-    if ctx.rule.kind == "ros2_interface_library":
+    if Ros2InterfaceInfo in target:
```
woutervdstoel: (This is also related to https://github.com/mvukov/rules_ros2/pull/617/changes#r3015105293)

woutervdstoel:
> There should be decent disclaimer about this. Instead of silently failing, what can be done to ensure proper error handline? -- not for this PR but I am curious.

I don't think we can handle this at analysis time since it requires looking at the content of the files. We could of course generate tests to assert it. I think that since it's a feature for advanced users anyway, maybe it doesn't need the greatest error handling?

> Also, what's the motivation to use .idl directly?

In our setup: we want to annotate certain message fields with some custom annotations that another tool looks as (which reuses the existing rosidl parser). The IDL format is much more flexible in this regard than the msg format.

I also happen to know that Apex.AI recommends their customers directly write IDL because of the extended feature set.

ros2/interfaces.bzl
```diff
@@ -47,11 +49,11 @@ def _ros2_interface_library_impl(ctx):
 
 ros2_interface_library = rule(
     attrs = {
+        "deps": attr.label_list(providers = [Ros2InterfaceInfo]),
```
mvukov: Please revert all cosmetic changes: here and below to make the diff minimal.

ros2/interfaces.bzl
```diff
@@ -436,6 +478,20 @@ c_generator_aspect = aspect(
     implementation = _c_generator_aspect_impl,
     attr_aspects = ["deps"],
     attrs = {
+        "_c_copts": attr.string_list(
```
mvukov: Here as well. If needed there can be a separate refactor PR.

ros2/interfaces.bzl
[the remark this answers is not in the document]
mvukov: Please remove from this PR, let's discuss in a follow-up PR.

ros2/interfaces.bzl
```diff
@@ -633,20 +689,10 @@ cpp_generator_aspect = aspect(
...
-            ],
-            providers = [CcInfo],
-        ),
-        "_cc_toolchain": attr.label(
-            default = Label("@bazel_tools//tools/cpp:current_cc_toolchain"),
+        "_typesupport_templates": attr.label(
+            default = Label("@ros2_rosidl_typesupport//:rosidl_typesupport_generator_cpp_templates"),
         ),
     },
-    required_providers = [Ros2InterfaceInfo],
```
mvukov, line 775: Let's discuss this in a separate PR.

ros2/interfaces.bzl
woutervdstoel: My autoformatter accidentally ran here, undid the changes

ros2/interfaces.bzl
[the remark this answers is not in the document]
woutervdstoel: Done

ros2/plugin_aspects.bzl
```diff
@@ -99,11 +99,12 @@ def create_interface_struct(target):
     return struct(
         package_name = target.label.name,
         srcs = target[Ros2InterfaceInfo].info.srcs,
+        generated_idl_files = target[IdlAdapterAspectInfo].generated_idl_files if IdlAdapterAspectInfo in target else None,
     )
 
 def _ros2_interface_collector_aspect_impl(target, ctx):
     direct_interfaces = []
-    if ctx.rule.kind == "ros2_interface_library":
+    if ctx.rule.kind == "ros2_interface_rule":
```
woutervdstoel:
```suggestion
    if ctx.rule.kind == "ros2_interface_library":
```

ros2/plugin_aspects.bzl
```diff
@@ -125,6 +126,7 @@ ros2_interface_collector_aspect = aspect(
     implementation = _ros2_interface_collector_aspect_impl,
     attr_aspects = _ROS2_COLLECTOR_ATTR_ASPECTS,
     provides = [Ros2InterfaceCollectorAspectInfo],
+    required_aspect_providers = [IdlAdapterAspectInfo],
```
woutervdstoel, line 134: This is necessary for the logic above to correctly be able to access the IdlAdapterAspectInfo

mvukov:
> I don't think we can handle this at analysis time since it requires looking at the content of the files. We could of course generate tests to assert it. I think that since it's a feature for advanced users anyway, maybe it doesn't need the greatest error handling?

For starters, I think it would be enough to clearly document this e.g. in ros2_interface_library rule docs. In a follow-up a proper build time testing can be introduced.

mvukov:
@woutervdstoel Any plans to wrap up this work?

woutervdstoel:
@mvukov I just pushed the documentation change you requested  - please let me know if there's anything else that still needs improvement

ros2/interfaces.bzl
```diff
@@ -47,11 +49,26 @@ def _ros2_interface_library_impl(ctx):
 
 ros2_interface_library = rule(
     attrs = {
+        "deps": attr.label_list(providers = [Ros2InterfaceInfo]),
```
mvukov: Here and elsewhere: please revert cosmetic changes of reorganizing attributes in this PR.

ros2/test/interfaces/py_messages_tests.py
```diff
@@ -0,0 +1,18 @@
...
+from test_messages.msg import Foo
+
+
+def test_foo():
+    foo = Foo()
+    assert foo.bar.baz.value == 0
+    foo.bar.baz.value = 42
+    assert foo.bar.baz.value == 42
+    with pytest.raises(AttributeError):
+        # This should raise an error because baz is not a member of Foo
```
mvukov: Please end sentence with a period.

ros2/test/interfaces/cpp_messages_tests.cc
```diff
@@ -0,0 +1,47 @@
+// Copyright 2026 Wouter van der Stoel
```
mvukov, line 1: The file name should end with `_tests.cc`.

ros2/test/interfaces/py_messages_tests.py
```diff
@@ -0,0 +1,18 @@
+import sys
```
mvukov, line 1: The file name should end with `_tests.py`.

ros2/interfaces.bzl
```diff
@@ -47,11 +49,26 @@ def _ros2_interface_library_impl(ctx):
...
         "srcs": attr.label_list(
-            allow_files = [".action", ".msg", ".srv"],
+            allow_files = [".action", ".idl", ".msg", ".srv"],
+            doc = """
+IDL files to generate code from.
```
mvukov, line 54: The docs is spot on.

mvukov (commented):
Left a few smaller comments. Please address them and then we can merge this PR. I'll proceed with finishing up Jazzy (@lalten FYI) afterward.

ros2/test/interfaces/cpp_messages_tests.cc
lalten: There should be a pre-commit hook to disallow certain filename regexes, this would speed up the review cycles then

lalten: filed https://github.com/mvukov/rules_ros2/issues/636

mvukov: Fixed in https://github.com/mvukov/rules_ros2/pull/637.

ros2/interfaces.bzl
```diff
@@ -104,13 +132,33 @@ def _run_adapter(ctx, package_name, srcs):
...
     idl_tuples = []
-    for src in srcs:
+
+    for src in idl_srcs:
+        idl_submodule_name = _get_idl_submodule_name(src)
+        stem = _get_stem(src)
+
+        new_file = ctx.actions.declare_file("{}/{}/{}.idl".format(package_name, idl_submodule_name, stem))
+        idl_files.append(new_file)
+        ctx.actions.run_shell(
```
mvukov: Would a symlink suffice here?

ros2/interfaces.bzl
```diff
@@ -88,10 +103,23 @@ def _to_snake_case(not_snake_case):
 def _get_stem(path):
     return path.basename[:-len(path.extension) - 1]
 
+def _get_idl_submodule_name(path):
```
mvukov: On second thought... `submodule` is misleading and not readable. I'd rename submodule -> type because this essentially determines a type (action, msg, srv).

ros2/interfaces.bzl
```diff
@@ -88,10 +103,23 @@ def _to_snake_case(not_snake_case):
...
+    """For .idl files, we do not know if they are services, actions or messages, so we enforce the convention
+    where the parent directory of the .idl file is the submodule name.
+    For .msg/.srv and other files, return the extension
+    """
+    if path.extension == "idl":
+        return path.dirname.split("/")[-1]
```
mvukov: If anything other than action/msg/srv this should error.

ros2/interfaces.bzl
```diff
@@ -125,29 +173,36 @@ def _run_adapter(ctx, package_name, srcs):
...
         executable = ctx.executable._adapter,
         arguments = [adapter_cmd_args],
         mnemonic = "Ros2IdlAdapter",
         progress_message = "Generating IDL files for %{label}",
     )
 
-    return idl_files, idl_tuples
+    return idl_files, generated_idl_files, idl_tuples
 
 IdlAdapterAspectInfo = provider("TBD", fields = [
```
mvukov, line 188: TBH, I don't see a need to touch this. But I'd like to learn more.

ros2/interfaces.bzl
```diff
@@ -104,13 +132,33 @@ def _run_adapter(ctx, package_name, srcs):
     output_dir = adapter_map.dirname
 
     idl_files = []
+    generated_idl_files = []
```
mvukov, line 139: On second thought: why don't you dump everything in idl_files?

IIUC, generated_idl_files is needed such that we put idl files to the ament index. We can dump all idl_files to the index IMO.

ros2/ament.bzl
```diff
@@ -213,6 +213,20 @@ def _ros2_ament_setup_rule_impl(ctx):
             outputs.append(src_file)
             idl_manifest_contents.append(paths.join(subdir, src.basename))
 
+        # Generated IDL files in the output tree such that the rosbag mcap writer can include them in the metadata.
+        # mcap supports both IDL and msg definitions, but not combinations. Therefore, if any handwritten IDL files are
+        # present, the rosbag writer also needs any generated IDL files it depends on.
+        if idl.generated_idl_files != None:
```
mvukov: IIUC, here you dump generated IDL files but not user-provided IDL files, right? How then e.g. rosbag knows how to handle user-provided IDL files?

mvukov (commented):
On second thought, I have more questions.

Please let me know if you have time to work on this in the coming days -- let's see whether makes sense to iron out this PR or I can proceed with #558 .

BTW, I pushed a small fix to your branch recently today.

mvukov (changes requested):
Until we sort out some details I pointed out earlier, I'll just make this PR as a not yet ready to be merged.

ros2/ament.bzl
woutervdstoel: The user-provided ones are also in `srcs`. I just pushed a fix to not distinguish user-provided vs generated idl files anymore.

ros2/interfaces.bzl
woutervdstoel: I had a reason for not doing it like that initially, but it seems to work right now.

ros2/interfaces.bzl
woutervdstoel: Yes, works

woutervdstoel:
@mvukov please have another look, I got rid of the distinction between generated_idl_files and idl_files

ros2/interfaces.bzl
```diff
@@ -125,29 +175,34 @@ def _run_adapter(ctx, package_name, srcs):
...
         progress_message = "Generating IDL files for %{label}",
     )
 
-    return idl_files, idl_tuples
+    return idl_files + generated_idl_files, idl_tuples
 
 IdlAdapterAspectInfo = provider("TBD", fields = [
     "idl_files",
     "idl_tuples",
+    "idl_deps",
```
mvukov: There's no need for this. The interface collector aspect takes care of idl_files. This field and it's assignment can be removed.

ros2/plugin_aspects.bzl
```diff
@@ -100,6 +100,7 @@ def create_interface_struct(target):
     return struct(
         package_name = target.label.name,
         srcs = target[Ros2InterfaceInfo].info.srcs,
+        idl_files = target[IdlAdapterAspectInfo].idl_files if IdlAdapterAspectInfo in target else None,
```
mvukov: If Ros2InterfaceInfo is required to be present in the target, IdlAdapterAspectInfo should be therein as well. It's what you request in required_aspect_providers.

Also, I'd return an empty list here.

ros2/ament.bzl
```diff
@@ -213,6 +213,20 @@ def _ros2_ament_setup_rule_impl(ctx):
...
+        if idl.idl_files != None:
+            for idl in idl.idl_files:
+                src_file = ctx.actions.declare_file(
+                    paths.join(prefix_path, "share", package_name, idl.dirname.split("/")[-1], idl.basename),
+                )
+                ctx.actions.symlink(
+                    output = src_file,
+                    target_file = idl,
+                )
+                outputs.append(src_file)
```
mvukov, line 231: Please also add files to the manifest to be consistent with handling of idl.srcs.

On a separate note, not necessarily for this PR: now that we dump idl.idl_files, do we need to dump idl.srcs to the ament index as well? As-is with this PR, the index will have redundant information (e.g. msg files and their correspondent .idl files).

I guess for rosbag/mcap idl files should be sufficient. But, we should check how this plays with e.g. foxglove bridge -- i.e. whether the bridge impl requires .action, .msg, .srv files.

ros2/plugin_aspects.bzl
woutervdstoel: IdlAdapterAspectInfo isn't always present because there are also plugins that don't have idl definitions associated.

The `required_aspect_providers` only constrains the order in which the aspects are applied (and makes the output of the first aspect available to the next) in case both apply. Despite the name, it doesn't enforce that the other Info is actually present.

ros2/interfaces.bzl
woutervdstoel: You're right, this is unrelated - it should be part of another internal patch for our repo

ros2/ament.bzl
woutervdstoel: Not all tooling supports reading `idl` definitions, so I'd say it's still preferable to have the msg definitions as well. Even rosbag/mcap prints a warning in case there's only IDL definitions available.

lalten:
bump on this :)

ros2/plugin_aspects.bzl
```diff
@@ -102,6 +102,7 @@ def create_interface_struct(target):
     return struct(
         package_name = target.label.name,
         srcs = target[Ros2InterfaceInfo].info.srcs,
+        idl_files = target[IdlAdapterAspectInfo].idl_files if IdlAdapterAspectInfo in target else [],
```
mvukov: One outstanding thing I gave some thoughts about: IdlAdapterAspectInfo is triggered only during code-generation on dep-graph nodes/vertices that have Ros2InterfaceInfo. Otherwise, IdlAdapterAspectInfo is never triggered.

For the sake of completeness, the following should be documented: if one just includes a ros2_interface_library target as a data dep on a target on which we want to apply ros2_ament_setup (via `deps`, directly or indirectly), no IDL files are going to be generated (as no code-generation happens). In other words, one needs to have some lang-bindings included for IDL files to be generated. I didn't make any tests to confirm this, but thought a bit deeper about this.

Sounds reasonable?

I realize that in most cases folks are going to have lang-bindings generated, but... for the sake of completeness this should be documented IMO.

mvukov:
Plus, the build needs to be fixed for Jazzy. Can you take care of this, please?

ros2/plugin_aspects.bzl
woutervdstoel: I agree that the aspect-based code generation comes with some quirks that may cause incomplete ament setups. But I don't think this PR changes anything about that, right? Wouldn't such documentation belong in a separate PR?

mvukov:
@woutervdstoel Brother Claude added a failing test and then a fix. Then the docs were updated. PTAL and lemme know if it makes sense. I'll merge the PR then.

woutervdstoel:
Yes, the new test and doc make sense to me. I think I ended up with `None` instead of the empty list because of the same problem that Claude fixed in a different way.

mvukov (approved):

PR #618, opened 2026-03-27
Remove -latomic from linkopts

linzhp:
-latomic is needed on 32-bit ARM (armv7), where the compiler cannot inline 64-bit atomic operations as native instructions and must call out to GCC's libatomic runtime. On      
  x86_64 and aarch64 — the targets rules_ros2 supports — the CPU provides native atomic instructions and the compiler always inlines them, making -latomic unused at runtime.      
                                                                  
  Removing it has two benefits:                                                                                                                                                    
  - Fixes compatibility with `zig cc` (used by as [hermetic_cc_toolchain](https://github.com/uber/hermetic_cc_toolchain/)) that do not bundle libatomic as a system library, causing link failures when -latomic is
  present                                                                                                                                                                          
  - Removes an unnecessary link-time dependency on platforms where it has no effect
                                                                                                                                                                                   
  If armv7 support is ever needed, -latomic can be reintroduced conditionally with select({"@platforms//cpu:armv7": ["-latomic"], "//conditions:default": []}).

### DIFF
--- a/repositories/iceoryx.BUILD.bazel
+++ b/repositories/iceoryx.BUILD.bazel
@@ -48,7 +48,6 @@ cmake(
         {
             ("@platforms//os:linux", "@platforms//os:macos"): [
                 "-lacl",
-                "-latomic",
                 "-lpthread",
                 "-lrt",
             ],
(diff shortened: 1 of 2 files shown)


lalten:
this would be great to have! @mvukov could you please retrigger CI?

linzhp:
Looks like the system Clang still need `-latomic`. So this PR won't work. On the other hand, latest Zig already ignores `-latomic` automatically: https://github.com/ziglang/zig/blob/738d2be9d6b6ef3ff3559130c05159ef53336224/src/target.zig#L338

So this PR is not needed after people upgrade Zig:

PR #625, opened 2026-04-29
Add missing imports attributes to py_library targets

linzhp:
Add missing imports attributes to py_library targets so aspect_rules_py's
pth generator adds the correct repo roots to sys.path. rules_python's
py_test launcher populates PYTHONPATH from all transitive sources, making
these importable even without explicit imports; aspect_rules_py's venv-based
py_test only emits .pth entries for targets with an explicit imports
attribute, so targets without one are invisible to the venv runner.


mvukov (approved):

mvukov:
BTW, please document the rationale for those extra imports at all places.

linzhp:
done

mvukov (approved):

PR #623, opened 2026-04-21
ament.bzl: Use marker files instead of subdirectories in packages resource index

furushchev:
Fixes #622 

- `Add _write_package_marker`: Create marker files at `packages/<name>` per ament specification
- Move `package.xml` to `share/<name>/package.xml` (correct location)
- Update `get_plugins_xml_path` to use share/ directory

**Before:**
  - share/ament_index/resource_index/packages/foo/package.xml (directory)

**After:**
  - share/ament_index/resource_index/packages/foo (marker file)
  - share/foo/package.xml (package manifest)

Fixes ament index spec compliance per:
https://github.com/ament/ament_cmake/blob/jazzy/ament_cmake_core/doc/resource_index.md

PR #628, opened 2026-05-08
feat: add ros2 interface CLI command support 

furushchev:
Add Bazel macro and Python wrapper to enable ros2 interface command
functionality within Bazel-built ROS 2 projects.

Changes:
- Add interface.bzl macro for creating ros2_interface binary targets
- Add ros2_interface.py Python wrapper exposing interface CLI verbs
- Register new files in ros2/BUILD.bazel exports_files
- Add `//chatter:interface` as an example usage
  - This example will fail on test without #623


ros2/ros2_interface.py
```diff
@@ -0,0 +1,33 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
mvukov, line 1: Add yourself as a copyright holder in relevant files.

mvukov (commented):
A general question: why/how is this utility useful in Bazel context? -- I am just curious.

Naming is going to be tricky here. We already have interfaces.bzl -- without proper docs (new) folks might be puzzled with interface.bzl and interfaces.bzl. WDYT?

furushchev:
> A general question: why/how is this utility useful in Bazel context? -- I am just curious.

In my Bazel workspace, ROS2 interface definitions are converted from other sources and are not in workspace as files, so it's worth porting `ros2 interface <command>` tool into the workspace to ensure the conversion tool works correctly.

> Naming is going to be tricky here. We already have interfaces.bzl -- without proper docs (new) folks might be puzzled with interface.bzl and interfaces.bzl. WDYT?

Currently, we have at least file level docstring both on `interface.bzl` and `interfaces.bzl`. Can you please clarify what you have in your mind about the proper docs in addition to them?

mvukov:
> > A general question: why/how is this utility useful in Bazel context? -- I am just curious.
> 
> In my Bazel workspace, ROS2 interface definitions are converted from other sources and are not in workspace as files, so it's worth porting `ros2 interface <command>` tool into the workspace to ensure the conversion tool works correctly.

I don't get this. IIUC, this tool can be used to list interfaces used for a specific target. Is this correct? Also, what kind of conversion you have in mind?

PR #631, opened 2026-05-12
cc_defs: Add -rdynamic to linkopts for clang compiler by default

### LINKED ISSUE #630: Pluginlib does not load plugins when compiled with LLVM toolchain
PluginLoader does not find declared plugins correctly when the libraries are compiled with LLVM toolchain, and ends with throwing an 
exception on calling `PluginLoader::createSharedInstance`.

To make the mechanism work with LLVM, executables must be linked with option `-rdynamic` (equivalent to `-Wl,--export-dynamic`)

To reproduce the error:

1. Add the following lines to `examples/MODULE.bazel`:

```diff
+bazel_dep(name = "llvm", version = "0.7.7")
+register_toolchains("@llvm//toolchain:all")
```

2. Run test for `image_pub_sub`

```bash
cd examples
bazel test //image_pub_sub/...
```

furushchev:
Closes #630 

This change adds default linker option `-rdynamic` to C++ binaries built with `ros2_cpp_binary`.
It's enabled only with LLVM toolchain (e.g. https://github.com/hermeticbuild/hermetic-llvm )

### DIFF
--- a/ros2/cc_defs.bzl
+++ b/ros2/cc_defs.bzl
@@ -3,6 +3,7 @@
 
 load("@com_github_mvukov_rules_ros2//ros2:ament.bzl", "sh_exec_launcher", "split_kwargs")
 load("@com_github_mvukov_rules_ros2//ros2:cc_opts.bzl", "C_COPTS")
+load("@com_github_mvukov_rules_ros2//ros2:cc_opts.bzl", "CPP_LINKOPTS")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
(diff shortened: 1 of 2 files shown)


mvukov:
Can we do this in a similar way as for clang? See https://github.com/mvukov/rules_ros2/blob/main/.bazelrc#L51-L53 -- we could have another llvm config there.

furushchev:
> Can we do this in a similar way as for clang? See https://github.com/mvukov/rules_ros2/blob/main/.bazelrc#L51-L53 -- we could have another llvm config there.

Hmm, this parameter is only necessary for binaries that link to the pluginlib library, so there's no need to set it globally. Therefore, I think it would be better to configure it in the current PR rather than writing it in `.bazelrc`.

mvukov:
This is llvm specific, so, I'd prefer to have this in an llvm config.

PR #633, opened 2026-05-22
fix: Resolve runfiles in launch_exec templates

furushchev:
The previous implementation used `$(rootpath)` to embed build-time paths to entry point executables. When launch wrappers were symlinked by other rules (such as `bazel_env`), these hardcoded paths became invalid, causing "file not found" errors at runtime.

To fix the issue:
- Replaced `$(rootpath)` with `$(rlocationpath)` in template substitutions
- Added Bazel runfiles library resolution in wrapper scripts:
  - `launch_exec.sh.tpl`: Use `rlocation` function
  - `launch_exec.py.tpl`: Use `python.runfiles.Runfiles.Rlocation`
- Added necessary dependencies:
  - `py_binary` / `py_test`: Add `@rules_python//python/runfiles` dependency
  - `sh_binary`/ `sh_test`: Enable `use_bash_launcher = True` for rlocation support


ros2/cc_defs.bzl
```diff
@@ -100,6 +100,7 @@ def _ros2_cpp_exec(target, name, ros2_package_name, set_up_ament, idl_deps, **kw
         name = name,
         srcs = [launcher],
         data = [target_impl],
+        use_bash_launcher = True,
```
mvukov, line 103: Does this mean that we had a bug before?

ros2/launch_exec.py.tpl
```diff
@@ -1,5 +1,6 @@
 import os
 import sys
+from python.runfiles import runfiles
```
mvukov, line 3: Please insert an empty line above as this is an external dep.

ros2/launch_exec.sh.tpl
```diff
@@ -12,10 +12,12 @@ if [ -n "${BAZEL_TEST:-}" ]; then
   fi
 fi
 
+ENTRYPOINT="$(rlocation {entry_point})"
```
mvukov, line 15: Don't you have to depend on a target from rules_sh for this to work? I remember that earlier versions of rules_sh required copy/pasting of a snippet.

mvukov (commented):
OK with me, left a few comments to be clarified before the merge.

PR #648, opened 2026-06-22
Add launch deps argument to ros2_launch

MerinoSheep:
Add a `launch_deps` argument to `ros2_launch` that accepts other
`ros2_launch` targets included by this launch file. Their runfiles are
propagated into this launch target, and their ROS 2 runtime deps are
included in the generated ament setup dependency collection.


mvukov:
This is interesting. Can you show some real-world use-cases for this new functionality?

MerinoSheep:
Yeah sure, In our current workflow, we have tiered launch.py files. With this PR, we no longer need to duplicate dependencies and targets across each layer.

mvukov:
> Yeah sure, In our current workflow, we have tiered launch.py files. With this PR, we no longer need to duplicate dependencies and targets across each layer.

Please illustrate this here somehow -- I don't see a big picture, yet.

PR #657, opened 2026-08-12
Dedup ros2_ament_setup outputs

ahans:
It can happen that ros2_ament_setup sees the same IDL deps multiple times (via different config transitions). In that case symlinks are created multiple times, leading to errors. Introduce a fix by deduping ros2_ament_setup's output.

### DIFF
--- a/.github/workflows/ci.bazelrc
+++ b/.github/workflows/ci.bazelrc
@@ -30,3 +30,6 @@ common --color=yes
 
 # Allows tests to run bazelisk-in-bazel, since this is the cache folder used
 test --test_env=XDG_CACHE_HOME
+
+# The default of 3 is not good enough, make Bazel try harder
+test --flaky_test_attempts=10
--- a/ros2/ament.bzl
+++ b/ros2/ament.bzl
@@ -195,6 +195,14 @@ def _ros2_ament_setup_rule_impl(ctx):
         ] + [idls_from_deps],
     ).to_list()
 
+    # Track generated and symlinked files to dedup outputs. It can happen that Ros2PluginCollectorAspectInfo sees a
+    # target such as builtin_interfaces multiple times via different edges of the dependency graph.  Those different
+    # paths can be introduced via config transitions (one concrete example is a C++ node with some IDL deps and a Python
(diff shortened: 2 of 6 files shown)


ahans:
@mvukov  I pushed another commit to this branch that gets rid of some of flakiness. I remember you removing the custom `ROS_DOMAIN_ID` in favor of the Bazel sandbox (#388). Unfortunately, that doesn't work inside a Docker container. I would have liked to bring back the `domain_coordinator` approach you had before, I think that'd be even more robust than random numbers. But at least the tests in this repo were sufficiently stable with just this. Pulling back in `ament_cmake_ros` for this seemed like too much effort. Happy to extract this into its own PR if you prefer.

ahans:
Trying to find out what introduced `linux-sandbox` being unavailable, I noticed that there is actually no container in the mix. So what was the reason for my local failures (using the provided devcontainer) is not the reason on CI. I came across https://github.com/bazelbuild/bazel/issues/24081, which points at the 24.04 upgrade as the cause, but when working locally on my 26.04 machine, `linux-sandbox` works just fine. So still not entirely sure what is going on. Either way, I think setting the DOMAIN_ID is a good extra line of defense in case network isolation doesn't work (for whatever reason).

mvukov:
:+1: Re:
* dedup: can it happen that files on different configs (via transitions) have different content? Off the top of my head, no, just wanted to hear your opinion as well.
* I'd prefer DOMAIN_ID is handled in a separate PR. In principle, both the devcontainer and the CI containers run in privileged mode:
  * https://github.com/mvukov/rules_ros2/blob/main/.devcontainer/ubuntu_24_04_x86/devcontainer.json#L15
  * https://github.com/mvukov/rules_ros2/blob/main/.github/workflows/main.yml#L54
  * In principle that should allow linux-sandbox, but I saw that in some occasions, it doesn't. Thanks for referencing the relevant Bazel issue.
  * Reverting removal domain_coordinator :+1: 

BTW, it annoyed me to have domain_coordinator (as-is in Python) because it introduces a non-trivial time overhead per test. Perhaps we can spend some tokens to have it in Rust/C++.

